PR #5, opened 2023-06-23
challenge-0-simpleNFT

technophile-04:
### Desc : 

**Completed with the functionalities** 

https://github.com/scaffold-eth/se-2-challenges/assets/80153681/a9255802-d48d-4acb-8299-b97da4b13aa1

I was wondering if should we go with SE-1-challenges horizontal card UI? 
Screenshot 2023-06-23 at 11 53 39 AM

----
~~Also, SE-1-chall has these two extra pages ? ~~
~~1. IPFS upload~~

~~2. IPFS download~~

~~Just wanted to confirm we should add them both right? Just wanted to confirm before I add some extra packages 🙌~~

I have added at [12620f7](https://github.com/scaffold-eth/se-2-challenges/pull/5/commits/12620f7ec1e8f9eab44f149056623887c8c11786) (check out demo video) but please let me know if we should remove them 🙌

----

### TODO : 
- [x] Cleanup and update UI a bit
- [X] 2 extra pages
- [X] ReadME (done with the basic iteration of README.md @Pabl0cks could you please take the look 🙏 )

PR #7, opened 2023-07-02
Base template update

technophile-04:
Added the learnings from #5


carletex:
Great stuff @technophile-04 

Thank you.

PR #8, opened 2023-07-03
Challenge-0-simple-nft

mertcanciy:
Added some information for beginners to describe what "yarn generate" command do


mertcanciy (commented):
Explanation for "yarn generate"

Issue #16, opened 2023-07-13
Implement Challenge 2 - Token vendor


rin-st:
- [ ] Contract
- [ ] UI
- [ ] Readme
- [ ] Tests

chnejohnson:
I'm very interested in this challenge.

Pabl0cks:
Closing this issue, we can create new issues for further iterations on Challenge 2. Good job!

PR #23, opened 2023-07-17
Challenge 2

rin-st:
Fixes https://github.com/scaffold-eth/se-2-challenges/issues/16

Frontend, contracts, tests, and deploying works as expected

Some questions:
1. Should I add Install node/yarn/git to readme or it should be only in the first challenge?
2. Should I remove gitpod?
3. Should I add that million console.logs from original scaffold-eth frontend?
4. Do we need usd equivalent for tokens? We show user balance in Eth by default, so it's strange to show that usd.
image
5. Should I show transfer/sell blocks if a user didn't buy tokens? Currently, I hide it.
6. Should I move part of readme before install to home page?

TODO:
- [x] Remove/comment Solidity code
- [x] Remove/comment Frontend code
- [x] Fix Todos


carletex:
Hey @rin-st! Thanks for working on this.

> Should I add Install node/yarn/git to readme or it should be only in the first challenge?

I think we can keep them. They do not take up much space, and it can be useful if someone skip some of the first challenges.

> Should I remove gitpod?

Yep, at least for now.

> Should I add that million console.logs from original scaffold-eth frontend?

haha. No. You can have some of them if you think they are useful.

> Do we need usd equivalent for tokens? We show user balance in Eth by default, so it's strange to show that usd.

Yep, maybe not. It looks a bit confusing.

> Should I show transfer/sell blocks if a user didn't buy tokens? Currently, I hide it.

Not a strong preference. Whatever you think is best.

---

This is not live yet, so happy to merge this even if is not perfect, and iterate over it.

rin-st:
Added some changes. I didn't remove solidity code for now, just commented. Probably we'll need it to fix things. Also, moved Vendor balances to the top block, so it becomes clear how tokens move.

If you need to test things, checkout to [e559ee1](https://github.com/scaffold-eth/se-2-challenges/pull/23/commits/e559ee1e56e01fc03a5cba57fce1c582f5b7e619) - commit before commenting code

Current version:
Снимок экрана 2023-07-18 в 14 51 46

Снимок экрана 2023-07-18 в 14 52 06

Снимок экрана 2023-07-18 в 14 53 15

carletex:
This is looking great @rin-st 

Maybe we can remove the contract code now.

@ZakGriffith is going to take a look (he has a lot of context with the SRE challenges) too.

rin-st:
Thanks @Pabl0cks ! Regarding tests, check `hardhat:test` script, it's the same. So it seems it's easier to change readme than duplicate scripts

carletex:
about the `test`command: 

`yarn hardhat:test` is the official for SE-2 (thinking that in the future we might have front-end tests) But I guess `yarn test` has more historical weight, so we added also it to the other challenges (could be seen like an alias?)

ZakGriffith:
I agree, this is solid and is very close to being complete. @Pabl0cks I see you're adding some changes to the PR, would you prefer we wait to merge this PR until those are complete?

Pabl0cks:
> I agree, this is solid and is very close to being complete. @Pabl0cks I see you're adding some changes to the PR, would you prefer we wait to merge this PR until those are complete?

I've just pushed my changes on the README, not going to make more changes for now. Can merge if you see them fine 👌

carletex:
Great job all!

Merging this since it looks pretty solid. Let's iterate on it in new PRs if needed.

PR #25, opened 2023-07-19
chall-1/fix wagmi cache issue on `deploy --reset`

technophile-04:
### Desc : 
It seems that the Staking UI was not updating when we were trying locally doing `yarn deploy --reset`

### Solution : 
used [useBlockNumber](https://0.12.x.wagmi.sh/react/hooks/useBlockNumber) to forcefully refetch all the red variable once new block is added. 

### Other solution explored : 
I tried all the options for `useContractRead` but none work (as we also encountered here -> https://github.com/scaffold-eth/scaffold-eth-2/pull/177#issuecomment-1435948027 (seems a bit hacky but found this to be minimal / working nicely)

PS: For SE-2-chall we are still using v12 wagmi and here are its docs -> https://0.12.x.wagmi.sh/

### Reason for the issue : 
Turns out the issue was related to wagmi cache problem, wagmi was using `cache` instead of refetching all the `read variables` (it does this until you hard refresh the page, or go to different tab and come back)

We already had some issues with wagmi `cache` in wagmi version less than `v1` checkout https://github.com/scaffold-eth/scaffold-eth-2/pull/177#issuecomment-1435948027  (but don't know if that still exists in `v1`)

### DIFF
--- a/packages/nextjs/pages/stakings.tsx
+++ b/packages/nextjs/pages/stakings.tsx
@@ -11,6 +11,7 @@ const Stakings: NextPage = () => {
     eventName: "Stake",
     fromBlock: 0,
   });
+  console.log("Stakevents", stakeEvents);
 
   if (isLoading)
     return (
@@ -47,9 +48,9 @@ const Stakings: NextPage = () => {
                   return (
                     <tr key={index}>
                       <td>
-                        <Address address={event.args.staker} />
+                        <Address address={event.args[0]} />
                       </td>
-                      <td>{event.args.amount && utils.formatEther(event.args.amount.toString())} ETH</td>
+                      <td>{event.args[1] && utils.formatEther(event.args[1].toString())} ETH</td>
                     </tr>
                   );
                 })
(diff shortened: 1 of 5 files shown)


packages/nextjs/pages/stakings.tsx
```diff
@@ -11,6 +11,7 @@ const Stakings: NextPage = () => {
     eventName: "Stake",
     fromBlock: 0,
   });
+  console.log("Stakevents", stakeEvents);
```
rin-st, line 14: do we need this log?

rin-st:
For some reason Total Staked doesn't update for me even after changes

rin-st:
There is a simpler solution: move all the logic to another component and make something like

```
const StakerUI: NextPage = () => {
  const { data: StakerContract } = useDeployedContractInfo("Staker");

return <Component key={StakerContract?.address} stakerAddress={StakerContract?.address}  />
```
It's almost the same as changing tabs.

Or if use changes you already made you can move only  total counter logic to another component, add key equal to address and it also works

rin-st (approved):

Pabl0cks:
Tested it and refresh is working great now! 🙌 

Added some minor tweaks to the README to not create a whole PR just for them, but are not related to your changes.

Pabl0cks (approved):

Issue #27, opened 2023-07-19
Challenge 1 Iteration Findings


ZakGriffith:
List of requests/bug findings from running challenge-1-decentralized-staking.  These are not set in stone so feel free to discuss.

**Feature Request** 
Is it possible to default the debug tab to land on the Staker.sol contract upon their initial visit?  

**Feature Request**
This is more of a general SE-2 discussion.  What about moving the Tx pop-up to be below the top tabs?
![image](https://github.com/scaffold-eth/se-2-challenges/assets/12072395/9e1b2d67-ec7c-439c-bb91-e3c211f21b86)

**Feature Request**
We need a more clear complete() process.  Currently all that changes is the total staked on the Staker UI tab as seen in the screenshot.  The old challenge had an atrocious bar that appeared at the top, so let's avoid it being that gross.  Possibly something that appears above the main Staker UI box displaying the amount staked?
![image](https://github.com/scaffold-eth/se-2-challenges/assets/12072395/c8d6009d-f5b3-4462-9965-bc8f04a992cd)

technophile-04:
Tysm @ZakGriffith this are great suggestions 🙌

> Is it possible to default the debug tab to land on the Staker.sol contract upon their initial visit?

Already fixed in #34

> This is more of a general SE-2 discussion. What about moving the Tx pop-up to be below the top tabs?
![Screenshot 2023-07-28 at 4 56 46 AM](https://github.com/scaffold-eth/se-2-challenges/assets/80153681/c1b3e8e4-9907-4056-b1fd-51be2c34d607)

Not sure about this, like in most apps usually they appear from top of the screen that's why it defaults in SE-2 but happy to happy update if others agree 🙌 

> We need a more clear complete() process. Currently, all that changes is the total staked on the Staker UI tab as seen in the screenshot. The old challenge had an atrocious bar that appeared at the top, so let's avoid it being that gross. Possibly something that appears above the main Staker UI box displaying the amount staked?

I didn't completely get it 😅 but are we talking about showing something more visual for the amount staked? something like this maybe? 
![Screenshot 2023-07-28 at 5 06 01 AM](https://github.com/scaffold-eth/se-2-challenges/assets/80153681/c7c85ffc-9dfa-4a92-94f1-eea976cb5708)

ZakGriffith:
> I didn't completely get it 😅 but are we talking about showing something more visual for the amount staked? something like this maybe?

[...]

PR #34, opened 2023-07-27
challenge 1 betterments

technophile-04:
## Description : 

**TODO's from #27 :**

- [X] Is it possible to default the debug tab to land on the Staker.sol contract upon their initial visit?  Fixed at fd50c9f0341e45d44bd25ce7358f666ec1532012

- [ ] This is more of a general SE-2 discussion.  What about moving the Tx pop-up to be below the top tabs?

- [x] We need a more clear complete() process.  Currently, all that changes is the total staked on the Staker UI tab as seen in the screenshot.  The old challenge had an atrocious bar that appeared at the top, so let's avoid it being that gross.  Possibly something that appears above the main Staker UI box displaying the amount staked?


technophile-04:
Added the UI to show Staked amount and ExternalContract triggered @[cd169cf](https://github.com/scaffold-eth/se-2-challenges/pull/34/commits/cd169cffd868d82e385be44851fd339d6091a92d) : 
Screenshot 2023-07-31 at 11 49 55 PM

Issue #35, opened 2023-07-28
Challenge 2 - Linting Errors


ZakGriffith:
index.tsx has a few linting and prettier errors/warnings that cause the yarn vercel command to fail unless ignored.

carletex:
In the base-template, we removed some checks to avoid newcomers having issues with types:

- Removed husky pre-commit hook
- Removed Github action

We decided we wanted to prioritize being friendly to beginners (starting with the challeges and getting block by types/eslint is frustating) over perfect types/eslint. They will have to face types/eslint on SE-2 :) (after the challenges)

**I think we should also remove the check when deploying.**  We'd just need to tweak `next.config.js` and ignoring typescript/eslint by default.

Thoughts? @ZakGriffith @technophile-04 @rin-st ?

technophile-04:
> I think we should also remove the check when deploying

Yup makes sense 🙌

rin-st:
Agree too

ZakGriffith:
Agreed as well

ZakGriffith:
Closing this as PR #39 ignores the linting errors, which allows us to keep the the quotes and apostrophes more readable.

PR #39, opened 2023-07-31
Ignore typescript/eslint errors on build

carletex:
See https://github.com/scaffold-eth/se-2-challenges/issues/35#issuecomment-1657808264

When merged, we should back-merge this to all the existing challenge branches


rin-st (approved):

carletex:
Back merged to challenges 0, 1, 2, 4 and 5

PR #42, opened 2023-08-01
Challenge 2 - Updates

ZakGriffith:
- General updates to the ReadMe
- Buy Events are now uncommented in events.tsx
- Vendor.sol no longer inherits Ownable, instructions added to ReadMe
- Divided vendor deploy script into sections to match the ReadMe

I plan to iterate on the ReadMe once again after merging since these changes were made in an IDE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ uint256 public constant tokensPerEth = 100;
 
 Edit `packages/hardhat/deploy/01_deploy_vendor.js` to deploy the `Vendor` (uncomment Vendor deploy lines).
 
+Uncomment the `Buy Tokens` sections in `packages\nextjs\pages\token-vendor.tsx` to show the UI to buy tokens on the Token Vendor tab.
+
 ### 🥅 Goals
 
 - [ ] When you try to buy tokens from the vendor, you should get an error: **'ERC20: transfer amount exceeds balance'**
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -96,6 +96,8 @@ uint256 public constant tokensPerEth = 100;
 
 Edit `packages/hardhat/deploy/01_deploy_vendor.js` to deploy the `Vendor` (uncomment Vendor deploy lines).
 
+Uncomment the `Buy Tokens` sections in `packages\nextjs\pages\token-vendor.tsx` to show the UI to buy tokens on the Token Vendor tab.
```
technophile-04, line 99: Should we also tell them to uncomment `tokenPerEth`? Or let them figure it out?

I mean both is fine for me and makes sense 🙌 just making sure we didn't forget to mention about it

technophile-04 (approved):
Awesome stuff, I don't remember SE-1 challenges had this description to uncomment frontend but I am loving it in SE-2-Challengs, hopefully because of this people will get familar with se-2 frontend features as well !!  Thanks 🙌

README.md
ZakGriffith: I think they should be able to figure it out.  Once the buyTokens section is uncommented, it will give them a nice error on tokensPerEth.  Let's let it ride like this and see how they handle it.

Issue #41, opened 2023-08-01
Challenge 2 - Updates


ZakGriffith:
- [x] In the packages/hardhat/deploy/01_deploy_vendor.js file, make a clear distinction between the vendor deploy, token transfer, and transfer of ownership sections.  
- [x] Vendor ETH and Token balances are commented out, with no note in the read me.  Either uncomment them or add instructions to the ReadMe.
- [x] Edit Vendor.sol to not inherit Ownable.sol.  Add instructions in the ReadMe to do so, including the code.
- [x] Uncomment the Buy Events section on the Events tab.  
- [x] Add a Sell Tokens Events section, commented, and have the user uncomment it via ReadMe instructions.
- [x] The Buy Tokens UI - We should either show it by default, or provide instructions in the ReadMe to uncomment.

ZakGriffith:
All except for the sell tokens events were completed in PR #42

ZakGriffith:
Last item completed in PR #43

PR #43, opened 2023-08-02
chall-2/add SellToken UI 

technophile-04:
### Descript : 

1 . Added the SellToken UI : 
   
Screenshot 2023-08-03 at 12 07 27 AM

2 . Updated the logic so that once the user uncomments we show Event Title  and loading instead of just showing loaders without titles : 
Screenshot 2023-08-02 at 11 58 41 PM

3 . Updated the ReadMe for the same 

tackles pt.5 of #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,9 @@ uint256 public constant tokensPerEth = 100;
 
 Edit `packages/hardhat/deploy/01_deploy_vendor.js` to deploy the `Vendor` (uncomment Vendor deploy lines).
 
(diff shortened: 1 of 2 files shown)


ZakGriffith:
Thank you, I like the style!  I am leaning toward having the buyTokens UI already uncommented from the start.  It doesn't cause any issues with the flow of the challenge, and shows "No Events Found" initially instead of a blank tab.  What do you think?

technophile-04:
> It doesn't cause any issues with the flow of the challenge, and shows "No Events Found" initially instead of a blank tab

Yeah I agree, also we tell them to add `BuyTokens` events at beginning of challenge so it makes sense 🙌

Updated in latest commits 🙌

ZakGriffith:
Nice!

Issue #37, opened 2023-07-28
Challenge 2 - Front end error after setting sepolia as targetNetwork


ZakGriffith:
When setting the targetNetwork in packages/nextjs/scaffold.config.ts to sepolia, I started getting a front end error when clicking into the Token Vendor tab.  When testing on localhost, below is the pop-up error received.

![image](https://github.com/scaffold-eth/se-2-challenges/assets/12072395/86c835a1-99c1-4544-9547-27d56370f17b)

When switching the targetNetwork to hardhat or goerli, there was no pop-up or console errors.

After deploying to vercel an error remained but only in the browser console.  The following is that error.

![image](https://github.com/scaffold-eth/se-2-challenges/assets/12072395/a4e36aeb-ee8b-48f9-b944-23b3db054f06)

I am unsure if this is due to the sepolia RPC acting up or not.

rin-st:
something wrong with sepolia rpc url and cors. Tried challenge 1 and met that error too but se-2 works well. Probably we need to update `wagmi` and it will be solved

technophile-04:
We recently merged #47 and back-merged it in every branch, there was a problem specifically with `sepolia` check out this discussion -> https://github.com/scaffold-eth/scaffold-eth-2/pull/381#issuecomment-1600306689 for more description 

But it should work now. Thanks Zak for mentioning it, closing this, for now, please feel free to reopen it if we still get the error 🙌

PR #47, opened 2023-08-07
fix: sepolia and wc2

rin-st:
fixes sepolia network error https://github.com/scaffold-eth/se-2-challenges/issues/37

copied https://github.com/scaffold-eth/scaffold-eth-2/pull/381

### DIFF
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -16,7 +16,7 @@
     "@ethersproject/networks": "^5.7.1",
     "@ethersproject/web": "^5.7.1",
     "@heroicons/react": "^2.0.11",
-    "@rainbow-me/rainbowkit": "^0.11.0",
+    "@rainbow-me/rainbowkit": "^0.12.15",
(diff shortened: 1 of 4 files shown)


technophile-04:
Tysm @rin-st 🙌, 

Regarding the CORS mentioned in #37 it goes away by this PR because we now actually use our own alchemy RPC provider instead of `https://rpc.sepolia.org` 

Actually `https://rpc.sepolia.org` has some problem with CORS I tried removing `alchemyProvider` from `wagmiConnectors`  in `SE-2` `main` and it gives CORS error there too 


So incase our alchmey RPC breaks and it fallbacks to sepolia public RPC its already broken too, this is present in SE-2 `main` too

Nothing we can do I think, I will create an issue on wagmi later but I found this discussion https://github.com/wagmi-dev/wagmi/issues/2802#issuecomment-1649737181

Nevertheless everything works great as we are using our own alchemy RPC starting this PR 🙌

PR #58, opened 2023-08-09
Challenge-4-dex

mertcanciy:
I created a simple sentence about that the user should check the required amount of tokens when they enter the ETH amount to the deposit function and according to the result, they can use approve() function


README.md
```diff
@@ -340,7 +340,7 @@ Let’s create two new functions that let us deposit and withdraw liquidity. How
 
 ### 🥅 Goals / Checks
 
-- [ ] 🥅 Before using the deposit function, check the required amount of $BAL tokens, which is equal to the amount of ETH you have entered for deposit by using the price function in debug tab.
+- [ ] 🥅 Before using the deposit function, check the required amount of $BAL tokens, which is equal to the amount of ETH you have entered for deposit by using the price function in debug tab. Now, you can use `approve()` function with only required amount of tokens.
```
rin-st, line 343: For me it's difficult to understand what do you mean.

As I understand, required amount is equal to `tokenDeposit = (msg.value * tokenReserve / ethReserve) + 1;`, isn't it? See `deposit()` function.
It's close to result of `price()`, but without fee

ZakGriffith: What about adding a side quest in the goals, something like this:

### 🥅 Goals / Checks

- [ ] 💧 Deposit liquidity, and then check your liquidity amount through the mapping in the debug tab. Has it changed properly? Did the right amount of assets get deposited?
> ⚔️ Side Quest:  Prove your knowledge by calculating the exact amount of Balloons you need to approve when depositing. Don't forget about the fee!
- [ ] 🧐 What happens if you `deposit()` at the beginning of the deployed contract, then another user starts swapping out for most of the balloons, and then you try to withdraw your position as a liquidity provider? Answer: you should get the amount of liquidity proportional to the ratio of assets within the isolated liquidity pool. It will not be 1:1.

rin-st: > What about adding a side quest in the goals, something like this:
> Deposit liquidity, and then

Good Idea, but again amount that needs to be approved before a deposit is still unclear if ratio is not 1:1.

rin-st:
@mertcanciy this pr should be merged with squash or even better from new branch. Please create new branches for new pr's

rin-st (changes requested):
create pr from new branch

README.md
mertcanciy: > For me it's difficult to understand what do you mean.
> 
> As I understand, required amount is equal to `tokenDeposit = (msg.value * tokenReserve / ethReserve) + 1;`, isn't it? See `deposit()` function. It's close to result of `price()`, but without fee

hmm I thought that the price exactly work as we use in ethToToken or tokenToEth functions, you are right. I think showing the token amount in deposit function could be really helpful for new users to make them understand the logic better. The point is that if we want them to use approve function with exactly correct amount of BAL tokens, we might need to consider about deleting the approving the high amount of tokens at the beginning of the challenge. I think the thing you said about approve amount yesterday was right ser @ZakGriffith . What do you think

mertcanciy:
> I've seen that sometimes pr are merged here without squash, so please create pr from new branch. Since we started to discuss, you can work on it here. Create a new branch and copy changes after work will be done.

I'm so sorry, I could not understand. Shall I create a new branch from challenge-4-dex branch here?

rin-st:
> I'm so sorry, I could not understand. Shall I create a new branch from challenge-4-dex branch here?

yes, but from origin [scaffold-eth:challenge-4-dex](https://github.com/scaffold-eth/se-2-challenges/tree/challenge-4-dex), not from your local branch. I'm afraid previous commits can be duplicated

README.md
rin-st: not sure how they need to calculate the exact amount, manually? and how to check if they're right? by allowance amount left after deposit?

mertcanciy: Yeah it can be done by getting the difference between the allowance before using deposit and after using the deposit function. Maybe by creating a pop-up that includes "... AMOUNTS OF $BAL WERE USED"

Or we can create a new function in smart contract for getting the token amount in deposit function just like the price function but specified for deposit.

Maybe we can find a better solution :)

mertcanciy:
> > I'm so sorry, I could not understand. Shall I create a new branch from challenge-4-dex branch here?
> 
> yes, but from origin [scaffold-eth:challenge-4-dex](https://github.com/scaffold-eth/se-2-challenges/tree/challenge-4-dex), not from your local branch. I'm afraid previous commits can be duplicated

Can you check if it is correct? https://github.com/scaffold-eth/se-2-challenges/pull/61

README.md
ZakGriffith: Here is a question, is there any real reason why someone would need to only approve the exact amount instead of an arbitrary larger number?  Maybe we're over thinking this.

Pabl0cks:
Not sure where to continue the discussion, so sending my thoughts first here to have all the previous comments. Been testing with 2:1 balance.

- @rin-st is correct with the calculations (`tokenDeposit = (msg.value * tokenReserve / ethReserve) + 1`)   
It has not much to do with price function, can be similar result for small deposits, but as soon as you increase the size of your deposit, the difference can be very big, since with price you calculate how are you going to unbalance the initial liquidity balance (as I understand).
- > Here is a question, is there any real reason why someone would need to only approve the exact amount instead of an arbitrary larger number?  

   I'm also having a hard time deciding if provides value to them to do the calculation.   
   We could just explain in the Readme to add a high enough quantity to not have problems of allowance. If we want, we could also add the Side Quest telling they can manually calculate the required allowance by checking `tokenDeposit` from `deposit()` function.
- Testing this I found a bug in `Events` tab, we got ETH/Balloons columns switched in Liquidity Provided and Liquidity Removed. I will create a PR to solve it.
- What I found a bit strange during my tests, is that when you Deposit Liquidity, we are taking the input as the ether we want to deposit, so we are actually just minting 2 liq if we deposit 4 ether in a 2:1 balance. But when we withdraw, we take the input as liquidity amount we want to withdraw, taking all the eth/balloons associated with that liquidity.   
Maybe is the normal behaviour, just felt a bit strange.

Just thinking out loud, would love to hear your thoughts @rin-st @ZakGriffith @mertcanciy 🙌

rin-st:
> We could just explain in the Readme to add a high enough quantity to not have problems of allowance. If we want, we could also add the Side Quest telling they can manually calculate the required allowance by checking tokenDeposit from deposit() function.

Agree, I think it's enough

> Testing this I found a bug in Events tab, we got ETH/Balloons columns switched in Liquidity Provided and Liquidity Removed. I will create a PR to solve it.

👍 

> What I found a bit strange during my tests, is that when you Deposit Liquidity, we are taking the input as the ether we want to deposit, so we are actually just minting 2 liq if we deposit 4 ether in a 2:1 balance. But when we withdraw, we take the input as liquidity amount we want to withdraw, taking all the eth/balloons associated with that liquidity.
Maybe is the normal behaviour, just felt a bit strange.

It feels a little bit strange, yes. But it difficult to think in terms of liquidity when you depositing - user need to calculate also how much Eth he will spend. For withdrawals it's not so critical, and he also can check his liq by `getLiquidity()`. Probably we need to change method name to `withdrawLiquidity()` to be more clear

README.md
rin-st: just to understand how it works 🤷 . Probably it's redundant, yes

mertcanciy:
>   We could just explain in the Readme to add a high enough quantity to not have problems of allowance. If we want, we could also add the Side Quest telling they can manually calculate the required allowance by checking `tokenDeposit` from `deposit()` function.

Yeah, I agree with you too. This can be enough

Pabl0cks:
> It feels a little bit strange, yes. But it difficult to think in terms of liquidity when you depositing - user need to calculate also how much Eth he will spend. For withdrawals it's not so critical, and he also can check his liq by getLiquidity(). Probably we need to change method name to withdrawLiquidity() to be more clear

You are right, it felt strange but it would probably get more confusing if we change how it works. 

We can change the method name or explain a bit more this piece of the Readme (or both!):

> 💬💬 More Hints: The withdraw() function lets a user take both ETH and $BAL tokens out at the correct ratio. The actual amount of ETH and tokens a liquidity provider withdraws could be higher than what they deposited because of the 0.3% fees collected from each trade. It also could be lower depending on the price fluctuations of $BAL to ETH and vice versa (from token swaps taking place using your AMM!). The 0.3% fee incentivizes third parties to provide liquidity, but they must be cautious of [Impermanent Loss (IL)](https://www.youtube.com/watch?v=8XJ1MSTEuU0&t=2s&ab_channel=Finematics). 

Maybe just changing first sentence to something like that?

> 💬💬 More Hints: The withdraw() function lets a user take his Liquidity Provider Tokens out, withdrawing both ETH and $BAL tokens out at the correct ratio.....

rin-st:
To sum up:
- remove side quest: `Side Quest: Prove your knowledge by calculating the exact amount of Balloons you need to approve when depositing. Don't forget about the fee!`
- explain in the Readme to add a high enough quantity to not have problems with allowance.
- change description of `withdraw` like it is in last @Pabl0cks comment
> 💬💬 More Hints: The withdraw() function lets a user take his Liquidity Provider Tokens out, withdrawing both ETH and $BAL tokens out at the correct ratio.....

@ZakGriffith @Pabl0cks @mertcanciy are you agree?

Pabl0cks:
> @ZakGriffith @Pabl0cks @mertcanciy are you agree?

Agree ser! Nice summary 🙌

mertcanciy:
> To sum up:
> 
> * remove side quest: `Side Quest: Prove your knowledge by calculating the exact amount of Balloons you need to approve when depositing. Don't forget about the fee!`
> * explain in the Readme to add a high enough quantity to not have problems with allowance.
> * change description of `withdraw` like it is in last @Pabl0cks comment
> 
> > 💬💬 More Hints: The withdraw() function lets a user take his Liquidity Provider Tokens out, withdrawing both ETH and $BAL tokens out at the correct ratio.....
> 
> @ZakGriffith @Pabl0cks @mertcanciy are you agree?

Agree, too!

Pabl0cks:
I've updated those changes in PR #61

Pabl0cks:
We can close this PR now, is solved with #61

Issue #63, opened 2023-08-09
Challenge 5 - Testing File


ZakGriffith:
Currently in Checkpoint 4, the testing file is not checking the Guru's balance when testing the `withdrawEarnings()` function.  It is only tracking the contract balance.  If that function withdraws the funds to a different account, it still passes the test file.  It would be beneficial for the Auto-grader if it made sure the funds were being sent to the correct address.

rin-st:
https://github.com/scaffold-eth/se-2-challenges/pull/76/commits/97096448063d907786e323c1152a95494977992b

technophile-04:
Was this solved?

rin-st:
yes, @ZakGriffith please check Challenge 5 tests

ZakGriffith:
Yup, fixed in #76

PR #76, opened 2023-08-27
Challenge 5 wagmi migration

rin-st:


rin-st:
Thank you for review @Pabl0cks , good points!

Details
> If I do a Cash Out Latest Voucher with Guru, the next cashout I try to do always throws this error, even if the first cashout was 0.05 and then there is only 0.02 additional to cashout (total 0.07):
Error: VM Exception while processing transaction: reverted with panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)

Couldn't reproduce it at all

> When I Close that channel with 0.95 ETH left, instead of getting 0.95 ETH back to the Rube, he gets 0.05 ETH.

How your channel has more than 0.5 ETH? It worked for me as expected as well

> If I have an open stream with a Rube, and then open a 2nd stream with a new Rube browser tab, the previous Rube browser tab shows like if there was no channel open. The same happends If i got the Guru stream section open, when I open the 2nd Rube channel and go back to the Guru, I see the 2 channels but then 1 dissapears (if I refresh, they all appear again)

Reason of the errors is because `useScaffoldEventSubscriber` listener is cached on first render and doesn't change after it. Fixed it. Difficult to debug btw.

> If I have "Streamer" section active with the Guru, and I open a channel with a Rube in another navigaror tab, the channel appears OK, but when I try to write any wisdom, I get this error:
TypeError: Cannot read properties of undefined (reading 'postMessage')

Same reason as in previous point

rin-st:
added an issue to the docs
https://github.com/scaffold-eth/se2-docs/issues/18

Pabl0cks:
Thanks for your review @rin-st ! 🙌 

> Couldn't reproduce it at all

> How your channel has more than 0.5 ETH? It worked for me as expected as well

Sorry for making you lose time there, maybe the problem is in my side of the streamer component/contract 😥. 

Could you share with me your resolved migrated challenge, so I can check the differences and fix my code to re-run the test again?

Thanks! ♥😊

rin-st:
I already added commit that fixed what I described. Pull changes from this branch and try again. Or it isn't fixed for you?

Pabl0cks:
> I already added commit that fixed what I described. Pull changes from this branch and try again. Or it isn't fixed for you?

This one got fixed for me:

> If I have an open stream with a Rube, and then open a 2nd stream with a new Rube browser tab, the previous Rube browser tab shows like if there was no channel open. The same happends If i got the Guru stream section open, when I open the 2nd Rube channel and go back to the Guru, I see the 2 channels but then 1 dissapears (if I refresh, they all appear again)

But still getting the error when I write the first Wisdom, and when I try to do a 2nd Cash Out. Plus now I'm getting a weird error before the "Time left" countdown gets to 0:

 `Error: VM Exception while processing transaction: reverted with reason string 'channel is not closing'
      at Streamer.timeLeft (contracts/Streamer.sol:45)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async HardhatNode.runCall (E:\GitHub\se-2-challenges\packages\hardhat\node_modules\hardhat\src\internal\hardhat-network\provider\node.ts:639:20)
      at async EthModule._callAction (E:\GitHub\se-2-challenges\packages\hardhat\node_modules\hardhat\src\internal\hardhat-network\provider\modules\eth.ts:354:9)`

So I think it might be problem of my Checkpoint 3 code + streamer contract, or I don't know 😥

rin-st:
updated pr. there was a problem with viem types, created issue on their side, link in last commit.

added checkpoint 3 solution, it should be deleted before merge

todo:
- [ ] revert solution commit

Pabl0cks:
> updated pr. there was a problem with viem types, created issue on their side, link in last commit.

With the last version almost everything fixed for me! Thanks Rinat 🙌 

Only this is still happening:

- If I have "Streamer" section active with the Guru, and I open a channel with a Rube in another navigaror tab, the channel appears OK, but when I try to write any wisdom, I get this error:
  `TypeError: Cannot read properties of undefined (reading 'postMessage')`
  <details>
  <summary>Screenshot</summary>
  
  image
</details>

**Note:** 
If I change the order of my actions, opening first the Rube sessions, and then go to Guru and go to "Streamer" section, I see the Rube channels opened and can write Wisdoms without error. Could it be some kind of initialization that is not doing when the tab is already open?

Pabl0cks (approved):
Working great for me with the last update Rinat! 🙌 Good job!! 🔥

rin-st:
Thank you for your help and patience @Pabl0cks !

Don't merge this for now, I need to revert commented blocks and checkpoint 3, and also update test file (but probably I'll fix tests in another pr). Will check it later today

packages/nextjs/pages/streamer.tsx
```diff
@@ -75,12 +75,21 @@ const Streamer: NextPage = () => {
...
     eventName: "Opened",
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    listener: (addr, _) => {
-      setOpened([...opened, addr]);
-      if (userIsOwner) {
-        setChannels({ ...channels, [addr]: new BroadcastChannel(addr) });
-      }
+    listener: logs => {
+      logs.map(log => {
+        const addr = log.args[0] as string;
```
technophile-04, line 80: I think @rin-st this is not working, for some reason this is not working and I need to destructure it make it work eg  :`log.args.user`. 

<details>
<summary> 
My Contract : 
</summary>

```
// SPDX-License-Identifier: MIT

pragma solidity 0.8.4;

import "@openzeppelin/contracts/access/Ownable.sol";
import "hardhat/console.sol";

error Streamer__User_already_running_a_channel();
error Streamer__Less_than_required_amount();
error Streamer__Signer_is_not_running_the_channel();
error Streamer__Transfer_failed();
error Streamer__User_channel_is_not_open();
error Streamer__User_channel_is_not_closed();
error Streamer__User_time_is_not_before_current_time();

/**
 * @title State Channel Application.
 * @author ABossOfMyself
 * @notice A minimal state channel application.
 * @dev Created a minimal state channel application, where users seeking a service lock collateral on-chain with a single transaction, interact with their service provider entirely off-chain, and finalize the interaction with a second on-chain transaction.
 */

contract Streamer is Ownable {
	/* ========== EVENTS ========== */

	event Opened(address user, uint256 amount);

	event Challenged(address user);

	event Withdrawn(address user, uint256 amount);

	event Closed(address user);

	/* ========== MAPPINGS ========== */

	mapping(address => uint256) balances;

	mapping(address => uint256) canCloseAt;

	/* ========== MUTATIVE FUNCTIONS ========== */

	function fundChannel() public payable {
		if (balances[msg.sender] != 0)
			revert Streamer__User_already_running_a_channel();

		if (msg.value < 0.5 ether) revert Streamer__Less_than_required_amount();

		balances[msg.sender] = msg.value;

		emit Opened(msg.sender, msg.value);
	}

	function timeLeft(address channel) public view returns (uint256) {
		if (canCloseAt[channel] == 0)
			revert Streamer__User_channel_is_not_closed();

		return canCloseAt[channel] - block.timestamp;
	}

	function withdrawEarnings(Voucher calldata voucher) public onlyOwner {
		// like the off-chain code, signatures are applied to the hash of the data
		// instead of the raw data itself

		bytes32 hashed = keccak256(abi.encode(voucher.updatedBalance));

		// The prefix string here is part of a convention used in ethereum for signing
		// and verification of off-chain messages. The trailing 32 refers to the 32 byte
		// length of the attached hash message.
		//
		// There are seemingly extra steps here compared to what was done in the off-chain
		// `reimburseService` and `processVoucher`. Note that those ethers signing and verification
		// functions do the same under the hood.
		//
		// again, see https://blog.ricmoo.com/verifying-messages-in-solidity-50a94f82b2ca

		bytes memory prefixed = abi.encodePacked(
			"\x19Ethereum Signed Message:\n32",
			hashed
		);

		bytes32 prefixedHashed = keccak256(prefixed);

		address signer = ecrecover(
			prefixedHashed,
			voucher.sig.v,
			voucher.sig.r,
			voucher.sig.s
		);

		if (balances[signer] <= voucher.updatedBalance)
			revert Streamer__Signer_is_not_running_the_channel();

		uint256 payment = balances[signer] - voucher.updatedBalance;

		balances[signer] -= payment;

		address owner = owner();

		(bool success, ) = owner.call{ value: payment }("");

		if (!success) revert Streamer__Transfer_failed();

		emit Withdrawn(owner, payment);
	}

	function challengeChannel() public {
		if (balances[msg.sender] == 0)
			revert Streamer__User_channel_is_not_open();

		canCloseAt[msg.sender] = block.timestamp + 30 seconds;

		emit Challenged(msg.sender);
	}

	function defundChannel() public {
		if (canCloseAt[msg.sender] == 0)
			revert Streamer__User_channel_is_not_closed();

		if (canCloseAt[msg.sender] > block.timestamp)
			revert Streamer__User_time_is_not_before_current_time();

		(bool success, ) = msg.sender.call{ value: balances[msg.sender] }("");

		if (!success) revert Streamer__Transfer_failed();

		balances[msg.sender] = 0;

		emit Closed(msg.sender);
	}

	struct Voucher {
		uint256 updatedBalance;
		Signature sig;
	}

	struct Signature {
		bytes32 r;
		bytes32 s;
		uint8 v;
	}
}

```


</details>

packages/nextjs/pages/streamer.tsx
```diff
@@ -141,8 +176,10 @@ const Streamer: NextPage = () => {
   useScaffoldEventSubscriber({
     contractName: "Streamer",
     eventName: "Challenged",
-    listener: addr => {
-      setChallenged([...challenged, addr]);
+    listener: logs => {
+      logs.map(log => {
+        setChallenged(challenged => [...challenged, log.args[0] as string]);
```
technophile-04, line 181: Same here

packages/nextjs/pages/streamer.tsx
```diff
@@ -164,8 +201,10 @@ const Streamer: NextPage = () => {
   useScaffoldEventSubscriber({
     contractName: "Streamer",
     eventName: "Closed",
-    listener: addr => {
-      setClosed([...closed, addr]);
+    listener: logs => {
+      logs.map(log => {
+        setClosed(closed => [...closed, log.args[0] as string]);
```
technophile-04, line 206: Same here

technophile-04:
Should we update this line in README.md the part regarding `ethers.utils.` ? : 

https://github.com/scaffold-eth/se-2-challenges/blob/af00b8249b9b99269d0a3d2da6419a7278afd78f/README.md?plain=1#L153 


Tysm Rinnat everything looking great !!, just added a small comment regarding event destructing 🙌

packages/nextjs/pages/streamer.tsx
rin-st: yes, if you have named event parameters, you need to destructure it like you said. But since base contract hasn't names in event parameters, I think we should leave it as is. And probably add this info to `useScaffoldEventSubscriber` docs

rin-st:
> Should we update this line in README.md the part regarding `ethers.utils.` ? :
> 
> https://github.com/scaffold-eth/se-2-challenges/blob/af00b8249b9b99269d0a3d2da6419a7278afd78f/README.md?plain=1#L153
> 
> Tysm Rinnat everything looking great !!, just added a small comment regarding event destructing 🙌

viem doesn't have splitSignature alternative for now. https://github.com/wagmi-dev/viem/discussions/458
So I think it's ok to not change it for now

rin-st:
fixed test from this issue
https://github.com/scaffold-eth/se-2-challenges/issues/63

packages/nextjs/pages/streamer.tsx
technophile-04: Ohh yup yup my bad 🙌

technophile-04 (approved):
Really looking solid Rinat, tysm !!

ZakGriffith (approved):
Did a full run through and this is working well.

rin-st:
Great, thank you for reviews!

PR #79, opened 2023-09-02
challenge 4 wagmi migration

technophile-04:
Noticed that there was `deployedContract.ts` already had a an abi so removed it


rin-st (approved):

rin-st:
Looks good! Thanks!

Pabl0cks (approved):
It's working great, gj Shiv! 🙌

Pushed small README tweaks from other migrations, will check other migrated Challenges in case those changes are missing ♥

PR #81, opened 2023-09-06
Challenge 1: Add README migration tweaks

Pabl0cks:
Modify burner wallet settings and modify `useScaffoldEventHistory` Hint to change `fromBlock: 3750241n` for BigInt support.


technophile-04 (approved):
Tysm Pablo 🙌 !!

PR #93, opened 2023-09-21
fix: stake events frontend

rin-st:
Events table doesn't work if when creating a contract, `Stake` event parameters aren't named `staker` and `amount`. In base Challenge contract params of `Stake` event has no names, also in [Readme](https://github.com/scaffold-eth/se-2-challenges/blob/challenge-1-decentralized-staking/README.md?plain=1#L9) too.

So we need to use [0] and [1] args by default.

### DIFF
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventHistory.ts
@@ -101,8 +101,16 @@ export const useScaffoldEventHistory = <
   ]);
 
   return {
-    data: events,
+    data: events?.map(addIndexedArgsToEvent),
     isLoading: isLoading,
     error: error,
   };
+};
+
+export const addIndexedArgsToEvent = (event: any) => {
+  if (event.args && !Array.isArray(event.args)) {
+    return { ...event, args: { ...event.args, ...Object.values(event.args) } };
+  }
+
+  return event;
 };
--- a/packages/nextjs/hooks/scaffold-eth/useScaffoldEventSubscriber.ts
+++ b/packages/nextjs/hooks/scaffold-eth/useScaffoldEventSubscriber.ts
@@ -1,7 +1,7 @@
 import { Abi, ExtractAbiEventNames } from "abitype";
 import { Log } from "viem";
 import { useContractEvent } from "wagmi";
-import { useDeployedContractInfo } from "~~/hooks/scaffold-eth";
+import { addIndexedArgsToEvent, useDeployedContractInfo } from "~~/hooks/scaffold-eth";
 import { getTargetNetwork } from "~~/utils/scaffold-eth";
 import { ContractAbi, ContractName, UseScaffoldEventConfig } from "~~/utils/scaffold-eth/contract";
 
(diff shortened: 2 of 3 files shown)


technophile-04:
Tysm @rin-st 🙌  

Actually, this has been bugging me since https://github.com/scaffold-eth/se-2-challenges/pull/76#discussion_r1315425663 😂, Its PITA that wagmi/viems `events.args` changes to array or object based on weather you have named parameter or not  (In `ethers` there was always to access events.args through array)

Like even though we have mentioned in README there  might be a case where use named parameters, and if they use named parameters things again break, what if we do this : 


```tsx
stakeEvents?.map((event, index) => {
  // evens.args can be array of values or object of values
  // if object then we convert to array of values
  if (event.args && !Array.isArray(event.args)) {
    event.args = Object.values(event.args);
  }
  return (
    <tr key={index}>
      <td>
        <Address address={event.args?.[0]} />
      </td>
      <td>{event.args?.[1] && formatEther(event.args?.[1])} ETH</td>
    </tr>
  );
})
```

This way irrespective of weather people use named parameters or not and also if they misspell parameter name we still show the events correctly ? What do you say? (We should also update this in other challenges if we like it 🙌)

Pabl0cks (commented):
> what if we do this instead ? : 
>
> ```tsx
> stakeEvents?.map((event, index) => {
>   // evens.args can be array of values or object of values
>   // if object then we convert to array of values
>   if (event.args && !Array.isArray(event.args)) {
>     event.args = Object.values(event.args);
>   }
>   return (
>     <tr key={index}>
>       <td>
>         <Address address={event.args?.[0]} />
>       </td>
>       <td>{event.args?.[1] && formatEther(event.args?.[1])} ETH</td>
>     </tr>
>   );
> })
> ```
>
> This way irrespective of weather people use named parameters or not and also if they misspell parameter name we still show the events correctly. What do you say? (We should also update this in other challenges if we like it 🙌)

In my case I was using a contract with `staker` and `amount` variable names, so I'm having trouble with Rinat's approach, but this with this tweak is working great, so I would vote to use this to cover all cases 🙌

rin-st:
good point! At first sight I thought Object.keys/values can be sorted alphabetically and it will break things. But everything works fine. In your example users can't use event.args.paramName. What about this?

useScaffoldEventHistory return:
```
return {
    data: events?.map(event => {
      if (event.args && !Array.isArray(event.args)) {
        return { ...event, args: { ...event.args, ...Object.values(event.args) } };
      }
      return event;
    }),
    isLoading: isLoading,
    error: error,
  };
```

so event args will be accessible both using indexes (always) and param names

rin-st:
Added as a temporary solution fix from https://github.com/scaffold-eth/scaffold-eth-2/pull/540 . Merging this since I think it blocks people from completing challenge 1. Types from se-2 pr doesn't work for this contract btw. 

Don't adding this to other challenges since there are no more mismatches between events naming in readme and destructuring on client. But will add it after se-2 pr is merged

rin-st:
created task to backmerge https://github.com/scaffold-eth/se-2-challenges/issues/95

PR #94, opened 2023-09-25
Challenge 2 token vendor

kmiguel10:


rin-st:
@kmiguel10 You don't need to push your solution here. This repo is about conditions, not solutions.

Just add a link of your solution repo to the submit modal on speedrunethereum.com

Thank you!

PR #97, opened 2023-09-30
stack roll events and dice roll content on small screen

pankaj512:
### Why
On small screen dice page shows roll event, winner event and game ui in single row, making events UI to overlap with center game UI. 

### What
On small screen dice page will show Game UI on top and then roll event and then winner event in stack form. 

### How
When screen size is small reduce the grid column to 1 and rearrange Game UI, Roll Event and Winner UI with grid's start-row property.

Screenshot from 2023-09-30 16-35-35

Screenshot from 2023-09-30 16-33-58


technophile-04:
Thanks @pankaj512  looks great 🙌

PR #98, opened 2023-10-01
Challenge 0 backmerge v2

rin-st:


Pabl0cks (approved):
Great job @rin-st !! I've been playing with the new version of challenge 0, and I think is working well 🙌 

Found a couple of issues that were already happening in the previous version:

- **When you send one of the NFTs from _"My NFTs"_ tab, it changes the order of the NFTs from the list, adding the last one of your NFTs in the spot of the NFT you just sent.** 

  Tried adding logging for that behaviour and seems to me (I might be very wrong here 😅) that `yourCollectibleContract.read.tokenOfOwnerByIndex` is grabbing the last NFT at `tokenIndex = 0`
  To fix it I've forced the order of `collectibleUpdate` array on 2afe84f831acbfcb8623e308a541234448059bdd.   
  If it isn't the correct approach, feel free to undo my change 🙌

- **When reproducing/testing the previous bug,  doing a lot of Mint/Send actions, at some point the `totalBalance` had some kind of cache problem, loading 2 different values.** 

  It was loading the correct one and another one bugged, for example 4 (correct) and 7 (maybe the value from the original `totalBalance` if I didn't send the NFTs?).  
  The list of NFTs loaded once with each`totalBalance` value, one correctly and one that throwed this exception:
  
  ```
  Error: VM Exception while processing transaction: reverted with reason string 'ERC721Enumerable: owner index out of bounds'
  ```
  
  Could not reproduce the exact pattern to make the 2nd bug happend, seemed pretty random. For the first bug I took those videos:
  
  Prefix:
  send-nft-refresh-list-prefix
  Postfix (in this case the random exception appeard):
  send-nft-refresh-list-postfix

technophile-04:
Tysm Rinat !! 

> When reproducing/testing the previous bug, doing a lot of Mint/Send actions, at some point the totalBalance had some kind of cache problem, loading 2 different values.

I actually really tried reproducing it but didn't happen to me but I am sure this might happen in some cases I just remove `cacheOnBlock`at [c00a350](https://github.com/scaffold-eth/se-2-challenges/pull/98/commits/c00a350a8404d80fa7b73bcec1582639c7a234b1)  it might be the culprit for this hopefully that fixes this 🤞

Merging this for no, Tysm guys 🙌

PR #101, opened 2023-10-01
feat: backmerge challenge 3

rin-st:


Pabl0cks (commented):
Impressive work with the backmerges on all challenges @rin-st 🙌

I'm having issues testing the Rigged Roll:
- It seems like every time you click on "_Rigged Roll_" calling the `riggedRoll` function 4 times.
- Trying the "_Rigged Roll_" for a lot of times, it's always getting the same roll number.

rin-st:
> Trying the "Rigged Roll" for a lot of times, it's always getting the same roll number.

fixed it. They changed api for timing, so you needed to wait 1000 seconds for next number 🤷‍♂️ . Sorry for not testing it properly, I tried it one time and thought it works as expected

> It seems like every time you click on "Rigged Roll" calling the riggedRoll function 4 times.

I have no idea why is it called 4 times. It's called one time from the frontend. Don't know how to fix this for now, will try to check again tomorrow.

@technophile-04 @carletex have you met something like that before?

technophile-04:
Tysm Rinat and Pablo for testing 🙌 !!  

> I have no idea why is it called 4 times. It's called one time from the frontend. Don't know how to fix this for now, will try to check again tomorrow.
> 
> @technophile-04 @carletex have you met something like that before?

Yup, will have a look 🙌

technophile-04:
> `DiceGame#rollTheDice` is called twice. Only last roll is registered/taking into account for prizes

Screenshot 2023-10-18 at 8 02 10 PM

Are you talking about these two calls 😅 ? 

- Actually  [`eth_call`](https://docs.alchemy.com/reference/eth-call) is just a simulation it does not cost any gas nor it's an actual transaction and `eth_sendRawTransaction` where actual transaction is sent. 

- Wagmi `useContractWrite` which internally uses `writeContract` which internally first calls [`simulateTransaction`](https://github.com/wagmi-dev/viem/blob/5c95fafceffe7f399b5b5ee32119e2d78a0c8acd/src/actions/public/simulateContract.ts#L118) which internally makes `eth_call`. So basically wagmi first make `eth_call` (i.e simulates transaction) if the transaction runs fully (i.e not revert or error out) then only it makes `eth_sendRawTransaction`. 

So that's why only the second call actually gets registered which is `eth_sendRawTransaction` which is the actual transaction. 

---


> RiggedRoll#riggedRoll is called 4 times. Only last one is counting for prizes.

Actually, after a lot of digging found that `wagmi` [retries `3 times`](https://viem.sh/docs/clients/transports/http.html#retrycount-optional) if any Json RPC calls fail (in our case it `eth_call`), So first it makes 1 call it fails and then it retires 3 times therefore `1+3 = 4`

We could fix this issue of making 4 times calls by adding `retryCount : 0`  fixed at [aef53a2](https://github.com/scaffold-eth/se-2-challenges/pull/101/commits/aef53a2e87bf28de5b510f493fff33a4c255aca1) 🙌


---

Also did some digging into why we are winning very very rarely : 

Screenshot 2023-10-18 at 10 10 08 PM

First of all lol the winning probability is a bit low by itself its 3/16 = 0.19 and on top of that if you look at the above image you can see `eth_call` succeeds it got `1` but `eth_sendRawTransaction` failed because it got `15` this is because b/w `eth_call` and `eth_sendRawTranaction` a block is mine which affects the hash and hence we get different result and transaction fails. So lol we rarely get the answer and sometimes even that fails since we are slow sending actual transactions and the block is mined b/w before it (**This case won't happen test network since blocks are mined a bit slow**)

Also in my notice lol getting a number less than `3` in hardhat environment seems very hardhat my wild and might be very wrong guess it because of maybe empty blockhashes which differ by only a very small amount (only variable I think there is `blockNumber` )


The easiest solution to this problem is simpling increasing the probability of winning maybe 5/16 = 0.3

Pabl0cks:
Awesome research @technophile-04 🤯

Tried last changes and it's working well to me, GJ Rinat and TY Shiv for the support! 🙌 

> So it seems that changing wins to 0..4 (chance to win 0,098 = 10%) or even 0..5 (0.14 = 14%) is reasonable.

I think it doesn't hurt to increase the chance. People will find their testing phase more smooth 👌

Pabl0cks (approved):

rin-st:
Experimented more with this.

> Wagmi useContractWrite which internally uses writeContract which internally first calls [simulateTransaction](https://github.com/wagmi-dev/viem/blob/5c95fafceffe7f399b5b5ee32119e2d78a0c8acd/src/actions/public/simulateContract.ts#L118) which internally makes eth_call. So basically wagmi first make eth_call (i.e simulates transaction) if the transaction runs fully (i.e not revert or error out) then only it makes eth_sendRawTransaction.

And `useContractWrite` and hence our `useScaffoldContractWrite` works for `eth_call`, not for `eth_sendRawTransaction` 😲 . You can check it by making some dices. If your first dice is successful (eth_call), but second is not (eth_sendRawTransaction), notification in our app says tx is successful, but no events fired. Also, we see error on hardhat logs, but no errors on frontend. So we see success notification and infinitely spinning dice 🤷‍♂️ .

Also tried to disable automining to include both `eth_call` and `eth_sendRawTransaction` to one block
```ts
mining: {
        auto: false,
        interval: 200,
      },
```
but it doesn't help, `eth_sendRawTransaction` mined in new block every time. So the same problems

technophile-04:
> Plus for some reason without automining useContractWrite and hence our useScaffoldContractWrite works for only for eth_call, not for eth_sendRawTransaction 😲

Ohh yeah strange, I tried adding `console.log` in `useTransactor` hook and it give me this : 

Screenshot 2023-10-23 at 7 45 12 PM


Notice its "Status" is  `reverted`  and [viem in their `waitForTransactionReceipt` docs says](https://viem.sh/docs/actions/public/waitForTransactionReceipt.html#waitfortransactionreceipt) : 

> If the Transaction reverts, then the action will throw an error.

But here it didn't throw an errror my wild guess is its something related to `useTransactor`, `pollingInterval`, `minningInterval` (lol 200ms and frontend polling interval defaults to 4_000 for hardhat/anvil I think)

--- 

> I think you noticed that notification error messages are different. The short one is for eth_call, long one is for eth_sendRawTransaction.

Oh I see, Also noticed that I think it happens only during using Custom solidity error because I tried with a normal revert statement and it gave me only one notification error box with my revert statement: 
Screenshot 2023-10-23 at 7 56 10 PM, we should handle it OG se-2 maybe too, I will try to create minimal reproducible example there 


Nevertheless, @rin-st I think we could merge this since this is not completely critical for the challenge and we can continue our tinkering in background 🙌

rin-st:
Updated readme and tests for 2 -> 5

Merging this, thanks for reviews!

PR #104, opened 2023-10-02
don't use custom light blue color for copy icon

pankaj512:
### Why
Currently copy icon of address component uses custom light blue color but this cause issue when some custom background is used. The issue is icon not clearly visible against some background.

### What
Change copy icon of address component to not use custom icon color but keep it default so that it auto adjust against background.

### How
Remove custom tailwind css text property.

Test method - add a new address component in debug tab under contact variable so see different against 3 different background.

#### Before Light
Screenshot from Screencast from 02-10-23 11:18:02 AM IST webm - 1

#### After Light
Screenshot from Screencast from 02-10-23 11:17:03 AM IST webm_1

#### Before Dark
Screenshot from Screencast from 02-10-23 11:17:51 AM IST webm - 3

#### After Dark
Screenshot from Screencast from 02-10-23 11:17:17 AM IST webm

### DIFF
--- a/packages/nextjs/components/scaffold-eth/Address.tsx
+++ b/packages/nextjs/components/scaffold-eth/Address.tsx
@@ -101,10 +101,7 @@ export const Address = ({ address, disableAddressLink, format, size = "base" }: 
         </a>
       )}
       {addressCopied ? (
-        <CheckCircleIcon
-          className="ml-1.5 text-xl font-normal text-sky-600 h-5 w-5 cursor-pointer"
-          aria-hidden="true"
-        />
+        <CheckCircleIcon className="ml-1.5 text-xl font-normal h-5 w-5 cursor-pointer" aria-hidden="true" />
       ) : (
         <CopyToClipboard
           text={address}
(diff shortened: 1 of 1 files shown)


rin-st:
Looks good! Thanks!

rin-st (approved):

Pabl0cks (approved):
Awesome job @pankaj512!! This dark mode icon color was catching my eye during my last tests and just realized you fixed it! ♥

@rin-st Do you think it could be interesting to merge this and backmerge to all the active se-2-challenges PRs so it's fixed on the next merge of each Challenge?  
Or it's easier to just change that couple of lines on each PR? I'm still a bit git-noob with merges/backmerges

technophile-04:
Thanks @pankaj512 for this !!

>  backmerge to all the active se-2-challenges PRs so it's fixed on the next merge of each Challenge

Yeah I think we can do it 🙌 switch PR branch and run `git merge base-challenge-template` should do the job but lol if there are many bad conflicts then your other suggestion makes more sense 

will try to backmerge this in all PR 🙌

PR #106, opened 2023-10-03
Challenge 4: Guard against vulnerability with the deposit function

### LINKED ISSUE #105: Challenge 4: Potential vulnerability with deposit function
I noticed that the provided solution code for the `deposit` function is transferring the $BAL tokens after updating the liquidity mapping. I think this may make the contract vulnerable to a reentrancy attack where a user could deposit ETH to be given liquidity without sending any $BAL. The malicious actor could then call the `withdraw` function and drain liquidity from the pool

image

MattPereira:
Fixes #105 

The provided solution code for the `deposit` function updates the `liquidity` mapping before transferring $BAL tokens to the DEX

I think it would be better to transfer the tokens from the caller of `deposit()` before updating the `liquidity` mapping in order to protect the DEX contract from being drained via reentrancy attack

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,11 +310,12 @@ Let’s create two new functions that let us deposit and withdraw liquidity. How
         // 💡 Discussion on adding 1 wei at end of calculation   ^
         // -> https://t.me/c/1655715571/106
 
+        require(token.transferFrom(msg.sender, address(this), tokenDeposit));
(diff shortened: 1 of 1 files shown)


MattPereira:
Just learned that contract state gets reset if a transaction reverts

PR #108, opened 2023-10-17
Fix test on token liquidity added to DEX

programmer-ke:
## What

Fixed the Liquidity provided test by adding 1 wei unit to the expected number of tokens deposited.

## Why

When calculating the tokens added when depositing liquidity, we add 1 wei as a safeguard. As the solution code demonstrates in Checkpoint 5 here: https://speedrunethereum.com/challenge/minimum-viable-exchange

```
        tokenDeposit = (msg.value * tokenReserve / ethReserve) + 1;
        // 💡 Discussion on adding 1 wei at end of calculation   ^
        // -> https://t.me/c/1655715571/106

```
However, adding 1 as above breaks the test that checks the `LiquidityProvided` event for the added tokens.

err

Fixed the test by adding one wei to the expected number of tokens deposited.

## Steps to Reproduce

- Clone the repo
- Checkout the branch `challenge-4-dex`
- Copy all the solution code from the README into DEX.sol
- Implement the `getLiquidity` getter method
- Run `yarn test`

The test will give the test error shown above.

Also reported in the DEX challenge group [here](https://t.me/c/1655715571/5065)

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ Let’s create two new functions that let us deposit and withdraw liquidity. How
         totalLiquidity += liquidityMinted;
 
         require(token.transferFrom(msg.sender, address(this), tokenDeposit));
-        emit LiquidityProvided(msg.sender, tokenDeposit, msg.value, liquidityMinted);
+        emit LiquidityProvided(msg.sender, liquidityMinted, msg.value, tokenDeposit);
         return tokenDeposit;
     }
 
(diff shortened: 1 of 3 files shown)


Pabl0cks (approved):
Nice catch @programmer-ke 🙌

The fix is looking good to me, CC @technophile-04 or @rin-st  for a quick second check. 

I've tested some changes in the Challenge 4 last week, but didn't catch the error because I copied the `DEX.sol` from a recent approved submission, I've just checked its code and had this on it:

`// Removing the + 1 for tokensDeposited as it doesn't pass the tests`

GJ again.

carletex:
**Note**: When changing a test on any challenge, we need to update that test in the autograder too. Currently, it's a manual process, so l lmk if we merge this.

PR #116, opened 2023-12-18
Challenge 3 dice game

Hotmanics:
Changes:

Tests
    1. Standardized the `rollAmount` across the entire test suite.
    ~~2.  Test `Should call diceGame.rollTheDice for a roll <= 5`~~
        ~~a. The `riggedRoll` function is now properly awaited.~~
        b. ~~The `riggedRoll` function now  properly gets passed in a `msg.value`.~~
        c. Removed nested `it` checks and moved their Chai checks to the end of the test.
        d. Updated the `Roll` event check to have the proper amount of parameters and values.

### DIFF
--- a/packages/hardhat/test/Challenge3.ts
+++ b/packages/hardhat/test/Challenge3.ts
@@ -85,44 +85,39 @@ describe("🚩 Challenge 3: 🎲 Dice Game", function () {
   });
 
   describe("🔑 Rigged Rolls", function () {
-    it("Should call diceGame.rollTheDice for a roll <= 5", async () => {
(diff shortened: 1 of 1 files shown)


technophile-04 (approved):
Thanks @Hotmanics, makes sense to me ! 

Nice catch that we were not testing the events emitted by diceGame contract. 

cc @ZakGriffith @rin-st or @Pabl0cks , I think we need to update autograder after we merge this 🙌 ..... also did we update autograder last time when we changed winning number from `<=2` to `<= 5` ?

rin-st (approved):

technophile-04:
Merging this, thanks all 🙌

Issue #115, opened 2023-12-13
challenge 6: update `backend-local` to use TS


technophile-04:
### Description : 

We are currently using a basic `.js` in `backend-local`  maybe we convert it to TS just for overall consistency, not super important

carletex:
Fixed on #122

PR #122, opened 2024-01-07
Update backend-local to use TS

rin-st:
closes https://github.com/scaffold-eth/se-2-challenges/issues/115


packages/backend-local/package.json
```diff
@@ -1,12 +1,21 @@
...
     "dotenv": "^16.3.1",
     "express": "^4.18.2"
+  },
+  "devDependencies": {
+    "@types/cors": "^2.8.17",
+    "@types/dotenv": "^8.2.0",
+    "@types/express": "^4.17.21",
+    "@types/node": "^20.10.6",
+    "nodemon": "^3.0.2",
+    "ts-node": "^10.9.2"
```
Pabl0cks, line 19: I was receiving this error when executing `yarn backend-local` => _Error: Cannot find module 'typescript'_

Fixed adding `"typescript": "^5.1.6"` dependency here.

Pabl0cks (commented):
After the dependency fix, tested the challenge and is working great for me! 

GJ @rin-st! 🙌

technophile-04 (approved):
Thanks @rin-st looking great, just pushed few changes thinking from perspective of deploying it to prod 🙌

Also it seems, `ts-node` has some issue with node 20 version if you use try to use `yarn backend-local` with node version `20` it will throw error `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"...` checkout this [stackoverflow discussion](https://stackoverflow.com/questions/76725253/ts-node-and-esm-unknown-file-extension-ts)

It seems people are suggestion to use [`tsx`](https://github.com/privatenumber/tsx) instead of `ts-node`, if we choose to use `tsx` we can also [remove nodemon](https://dev.to/rennycat/use-tsx-instead-of-nodemon-4lh3)

packages/backend-local/package.json
rin-st: I thought it's installed on the parent repo and it's enough, but I'm wrong. It seems I installed ts globally. And yes we need it in backend package too, thanks!

rin-st:
> Also it seems, ts-node has some issue with node 20 version if you use try to use yarn backend-local with node version 20 it will throw error TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"... checkout this [stackoverflow discussion](https://stackoverflow.com/questions/76725253/ts-node-and-esm-unknown-file-extension-ts)

> It seems people are suggestion to use [tsx](https://github.com/privatenumber/tsx) instead of ts-node, if we choose to use tsx we can also [remove nodemon](https://dev.to/rennycat/use-tsx-instead-of-nodemon-4lh3)

Thanks for suggestions!

I opened [issues of tsx](https://github.com/privatenumber/tsx/issues) and a little bit scared, not sure it stable enough.

But first answer in SO discussion you mentioned solved the problem for me. https://stackoverflow.com/a/76758731 . It works fine with node 20/18/16. Also we can add a flag `--no-warnings=ExperimentalWarning`, but I think we don't need it. Just to remember to check from time to time if the bug fixed.

What do you think?

technophile-04:
> I opened [issues of tsx](https://github.com/privatenumber/tsx/issues) and a little bit scared, not sure it stable enough.

>  Also we can add a flag --no-warnings=ExperimentalWarning, but I think we don't need it. Just to remember to check from time to time if the bug is fixed.

Yup, agree !! Tysm Rinat merging this  🙌!

Issue #117, opened 2023-12-20
challenge 6: update `backend-local` to use JSON file as db


technophile-04:
### Description : 

Currently, we are storing the backend state in a variable kind of hacky, I think it will be great if we use JSON file as DB. 

This won't make much of a difference while doing the challenge locally, but we are planning to connect this repo's challenge 6 `backend-local` to our box where the prod server will live and it's good to have some kind of persistent storage there. 

Some helper libraries :

1. https://www.npmjs.com/package/node-json-db
2. https://github.com/typicode/lowdb
3. https://www.npmjs.com/package/simple-json-db

rin-st:
If these libs works well with [next api](https://github.com/scaffold-eth/se-2-challenges/pull/112/commits/bce1fd7c661972db630ef5cce380344983eccbe5#diff-d3d790f0e0208bcf6de775a731e917d0887bab763b6081e0ae4ad90d0207ee9f), we don't even need prod server, isn't it? Every app will have separate db. Or am I wrong?
I'll check it next days

technophile-04:
I actually tried using `lowdb` 2months ago with next api route but vercel don't allow write operation to file when you deploy to vercel :(

carletex:
Yes, nextjs server functions on Vercel doesn't give you FS access. I think you'll have to do it with any kind of vercel storage (database, blob). 
https://vercel.com/docs/storage

I'm happy to deploy it to one of our boxes if that makes things easier for people.

you'all tell me!

rin-st:
I ~~experimented~~ spent countless hours fighting with cors, next api (new) bugs, vercel env vars bugs and built that mega app 🤦‍♂️ (see only app folder, underscoreds are for old pages api).

BE: https://github.com/rin-st/multisig-backend

changes needed to do in the challenge-6. https://github.com/scaffold-eth/se-2-challenges/compare/challenge-6-multisig...ch6-update-api?expand=1

KV hobby plan includes 30 000 requests per month and 3 000 per day and I think it's probably enough for now. https://vercel.com/docs/storage/vercel-kv/usage-and-pricing .

rin-st:
Added env example to BE repo. You can use your storage vars to test it locally or pm me for my

technophile-04:
> I experimented spent countless hours fighting with cors, next api (new) bugs, vercel env vars bugs and built that mega app 🤦‍♂️ (see only app folder, underscoreds are for old pages api).

[...]

PR #124, opened 2024-01-12
Add reference to our deployed backend url for chains different than Hardhat

Pabl0cks:
Commented [here](https://github.com/scaffold-eth/se-2-challenges/issues/117#issuecomment-1889601235), adding a reference to our deployed backend url when the Challenge is deployed to chains different than Hardhat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,12 +106,14 @@ yarn start
 
 > In a fourth terminal window:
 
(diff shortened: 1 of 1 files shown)


rin-st (approved):
Thanks!

PR #126, opened 2024-01-26
Base challenge next 14 migration

rin-st:
Base challenge template migration to next 14/other changes after last migration. Good luck reading it :D (probably testing is enough for most of the things, I mostly copied se-2 code)

In addition all se-2 changes, changed colors a bit, see last commits. For example I disabled hover for top right buttons since hover color is almost the same as background in dark mode.

Next steps:
- migrate this branch after https://github.com/scaffold-eth/scaffold-eth-2/pull/700

- add changes to other challenges


Pabl0cks (approved):
Awesome job @rin-st !!

Been testing it and is looking great to me 🙌 To test it I tried to add the Challenge 0 features and everything seems to be working fine! 

> In addition all se-2 changes, changed colors a bit, see last commits. For example I disabled hover for top right buttons since hover color is almost the same as background in dark mode.

Right now the hover effect is very hard to notice. We could try something like this, or any similar approach that can be noticed in both dark and light mode:  `hover:bg-black hover:bg-opacity-20`
If we can think of a more elegant way to reach it, would be great! 🙃

rin-st:
> Right now the hover effect is very hard to notice. We could try something like this, or any similar approach that can be noticed in both dark and light mode: hover:bg-black hover:bg-opacity-20
If we can think of a more elegant way to reach it, would be great!

Changed it only for dark mode, since for light mode black/20 is not looking good

technophile-04 (approved):
Tysm Rinat, tested and working great !! 

I think maybe we should backmerge this into other challenges I think that should be a bit easy hopefully since they are forked from this branch 🙌

rin-st:
> I think maybe we should backmerge this into other challenges I think that should be a bit easy hopefully since they are forked from this branch 🙌

Yes, you right. When I wrote first message I thought wagmi migration is fast and it's better to backmerge one time instead of two. But it's not. I'll do backmerges, thanks!

PR #130, opened 2024-02-01
Ch1 next14 migration

rin-st:
- [ ] fix tests
- [ ] figure out why reverts of withdraw/execute doesn't work (or it only for me?)


packages/hardhat/test/Challenge1.ts
```diff
@@ -88,23 +92,24 @@ describe("🚩 Challenge 1: 🥩 Decentralized Staking App", function () {
...
         it("Should redeploy Staker, stake, not get enough, and withdraw", async function () {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
           const [owner, secondAccount] = await ethers.getSigners();
 
           const ExampleExternalContract = await ethers.getContractFactory("ExampleExternalContract");
           exampleExternalContract = await ExampleExternalContract.deploy();
 
           const Staker = await ethers.getContractFactory("Staker");
-          stakerContract = await Staker.deploy(exampleExternalContract.address);
+          stakerContract = await Staker.deploy(exampleExternalContractAddress);
```
Pabl0cks, line 102: I was having issues passing the test in this point, was throwing always that it was already completed. Think the problem may be we are passing an old completed address as argument.

Fixed for me adding this before, to refresh with the right address:
`exampleExternalContractAddress = await exampleExternalContract.getAddress();`

My approach may be totally wrong, sorry if so!

Pabl0cks (commented):
Great job again with the Next14 migration on this one @rin-st 

Only left a comment with an issue I was having when passing the tests.

packages/hardhat/test/Challenge1.ts
rin-st: I don't know, it works for me 🤷‍♂️ . Changed it, also changed waiting time to 72 hours instead of one hour, since readme says to make it 72 hours. Without it tests also fails, didn't noticed it because I tested with 72 seconds. Thanks!

Pabl0cks: Now they're working great for me! 🙌 I had same issue with time, testing with low timeframes, nice catch!

Pabl0cks (approved):
Looks good to me! 

I think we could merge Challenges 0 to 2 when https://github.com/scaffold-eth/scaffold-eth-2/pull/712 is merged and we add the fix to 'base-challenge-template' and the 3 challenges

rin-st:
> I think we could merge Challenges 0 to 2 when https://github.com/scaffold-eth/scaffold-eth-2/pull/712 is merged and we add the fix to 'base-challenge-template' and the 3 challenges

Added last se-2 changes
https://github.com/scaffold-eth/se-2-challenges/pull/132

Pabl0cks:
Chal 1 is working great also with the last SE-2 features. GJ @rin-st!

PR #134, opened 2024-02-17
Challenge 3 next14 migration

rin-st:


Pabl0cks (approved):
TYSM @rin-st! Looks good to me, pushed a couple of small nitpicks that were not related to this PR.

I've detected a strange behavior (not related to this PR either, it happens also in prod). In rare situations when there is a winner event, it appears in "Winner Events" but it does not show in "Roll Events", until you refresh the page and it does appear. There is a coincidence that most of the times it happens with a Roll of 0, but does not happen everytime there is a 0 🤷‍♂️

rin-st:
Thanks for review and updates Pablo! 🙏 

> I've detected a strange behavior (not related to this PR either, it happens also in prod). In rare situations when there is a winner event, it appears in "Winner Events" but it does not show in "Roll Events", until you refresh the page and it does appear. There is a coincidence that most of the times it happens with a Roll of 0, but does not happen everytime there is a 0 🤷‍♂️

Good catch, thanks! Fixed it [728bac1](https://github.com/scaffold-eth/se-2-challenges/pull/134/commits/728bac1d686288f1172020b2b46a5472ae3f4db9)

PR #128, opened 2024-01-31
Challenge 0 next14 migration

rin-st:
Migration of challenge 0 to next 14 and last se-2 changes

TODO:
- [x] fix tests

PR #136, opened 2024-02-20
Revert "Revert "Merge pull request #128 from scaffold-eth/ch0-next14-migration""

carletex:
Reverts scaffold-eth/se-2-challenges#135

Will merge when the autograder is updated!


carletex:
Tested this branch with the new autograder and it works! Merging!

PR #137, opened 2024-02-21
Ch0: Update test for the autograder

carletex:
- Contracts are downloaded from the autograder and run locally.
- Had to add the prefix "download", if not getting this (typechain to blame, doesn't like the files starting with numbers.)

If doesn't affect the test workflow for the builder, we can merge it.

We need to do this for all the other challenges too. SE-1 tests did this too (e.g. https://github.com/scaffold-eth/scaffold-eth-challenges/blob/challenge-1-decentralized-staking/packages/hardhat/test/challenge_1.js), but not sure why/when we changed it in the SE-2.

### DIFF
--- a/packages/hardhat/test/Challenge0.ts
+++ b/packages/hardhat/test/Challenge0.ts
@@ -19,17 +19,20 @@ describe("🚩 Challenge 0: 🎟 Simple NFT Example 🤓", function () {
 
   describe("YourCollectible", function () {
     const contractAddress = process.env.CONTRACT_ADDRESS;
+
+    let contractArtifact: string;
     if (contractAddress) {
-      it("Should connect to external contract", async function () {
-        myContract = await ethers.getContractAt("YourCollectible", contractAddress);
-        console.log("     🛰 Connected to external contract", myContract.address);
-      });
+      // For the autograder.
+      contractArtifact = `contracts/download-${contractAddress}.sol:YourCollectible`
     } else {
-      it("Should deploy YourCollectible", async function () {
-        const YourCollectible = await ethers.getContractFactory("YourCollectible");
-        myContract = await YourCollectible.deploy();
-      });
+      contractArtifact = "contracts/YourCollectible.sol:YourCollectible";
     }
+
+    it("Should deploy the contract", async function () {
+      const YourCollectible = await ethers.getContractFactory(contractArtifact);
+      myContract = await YourCollectible.deploy();
+      console.log("\t"," 🛰  Contract deployed on", myContract.address);
+    });
 
     describe("mintItem()", function () {
       it("Should be able to mint an NFT", async function () {
--- a/packages/hardhat/test/Challenge0.ts
+++ b/packages/hardhat/test/Challenge0.ts
@@ -1,11 +1,5 @@
 //
-// this script executes when you run 'yarn test'
-//
-// you can also test remote submissions like:
-// CONTRACT_ADDRESS=0x43Ab1FCd430C1f20270C2470f857f7a006117bbb yarn test --network sepolia
-//
-// you can even run mint commands if the tests pass like:
-// yarn test && echo "PASSED" || echo "FAILED"
+// This script executes when you run 'yarn test'
 //
 
 import { ethers } from "hardhat";
--- a/packages/hardhat/test/Challenge0.ts
+++ b/packages/hardhat/test/Challenge0.ts
@@ -17,7 +17,7 @@ describe("🚩 Challenge 0: 🎟 Simple NFT Example 🤓", function () {
     let contractArtifact: string;
     if (contractAddress) {
       // For the autograder.
-      contractArtifact = `contracts/download-${contractAddress}.sol:YourCollectible`
+      contractArtifact = `contracts/download-${contractAddress}.sol:YourCollectible`;
     } else {
       contractArtifact = "contracts/YourCollectible.sol:YourCollectible";
     }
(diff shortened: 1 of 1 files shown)


rin-st:
Probably it was changed because at the top of the file we have
```
//
// this script executes when you run 'yarn test'
//
// you can also test remote submissions like:
// CONTRACT_ADDRESS=0x43Ab1FCd430C1f20270C2470f857f7a006117bbb yarn test --network sepolia
//
// you can even run mint commands if the tests pass like:
// yarn test && echo "PASSED" || echo "FAILED"
```
and it works with
`myContract = await ethers.getContractAt("YourCollectible", contractAddress);`
but doesn't work with 
```
// download- or not doesn't matter
contractArtifact = `contracts/download-${contractAddress}.sol:YourCollectible`;
const YourCollectible = await ethers.getContractFactory(contractArtifact);
myContract = await YourCollectible.deploy();
```
also noticed that for some reason network hardhat is hardcoded in test script of package.json
`"test": "REPORT_GAS=true hardhat test --network hardhat",`, so I needed to remove network to run network sepolia

technophile-04:
> and it works with
> myContract = await ethers.getContractAt("YourCollectible", contractAddress);
> but doesn't work with

Umm didn't get it properly, is that `await ethers.getContractFactory(contractArtifact)` is not working for you ? for me it runs correctly. 

> "test": "REPORT_GAS=true hardhat test --network hardhat"

Yup we should remove this, btw I think in autograder we are not passing `--network` flag while running tests. 


Btw found this gem if anyone wants to learn how autograder works internally not sure if its outdated though https://youtu.be/Rq0uW-lDngs?si=jM7GupcyDkQa6Mx-&t=1308

carletex:
Thanks for the review Rinat & Shiv!

> "test": "REPORT_GAS=true hardhat test --network hardhat",, so I needed to remove network to run network sepolia
> Yup we should remove this, btw I think in autograder we are not passing --network flag while running tests.

Yep, noticed that too! The autograder is currently passing the flag, but I was going to remove it since it was already hardcoded in the challenge. **So maybe we want to remove it from the challenges (if the `yarn test` works by default on hardhat) and roll it back in the autograder PR??**

https://github.com/austintgriffith/speedrun-grader/pull/9/files

----

OK, so if we want to keep the ability for people to test remote contracts we should have 3 options:

1. Local test (for speedrunners)
2. Remote test (for speedrunners)
3. Fetch-contract-local-test (for te autograder)

I could tweak the autograder as needed (like declaring a new env var just for the autograder: e.g. `DOWNLOADED_CONTRACT_ADDRESS`)

What do you guys think?

----

> Btw found this gem if anyone wants to learn how autograder works internally not sure if its outdated though https://youtu.be/Rq0uW-lDngs?si=jM7GupcyDkQa6Mx-&t=1308

Damn :sweat_smile: hahaha

technophile-04:
Actually we have hardcoded `--network hardhat` in main SE-2 repo too.

So if you don't pass flag it defaults to "defaultNetwork" set in `hardhat.config.js`.

So if we plan to remove the hardcoded `-network hardhat` which I am leaning too but lol not 100% sure since hardcoding it also kindof make sense (basically saving people from running test script on mainnet / testnet)

In case we plan to remove the hardcoded `--network hardhat` and this will also allow people to remote test contract

#### TODO:

1 . Specifically in challenge-0 README.md move `yarn test` above Deploying the contract section (Since in deploying contract section we are telling to update `defaultNetwork: "sepolia"`)

__In other PR__
2 . Update base-tempalte-challenge and other challenges to remove hardcoded `--network hardhat` flag

3 . Update SE-2 main repo

carletex:
> Actually we have hardcoded --network hardhat in main SE-2 repo too.
> So if you don't pass flag it defaults to "defaultNetwork" set in hardhat.config.js.

Oh oh, this makes sense.

So:

image

IDK, this looks like a lot of hassle for allowing testing a live contract (which I don't feel is super useful.... I personally have never done it.)

---

The alternative is to just merge this PR (+ change the top comment on the test file), right? Now I'm leaning more toward this... haha 


But still open to anything!

packages/hardhat/test/Challenge0.ts
```diff
@@ -19,18 +19,21 @@ describe("🚩 Challenge 0: 🎟 Simple NFT Example 🤓", function () {
...
-      });
+      // For the autograder.
+      contractArtifact = `contracts/download-${contractAddress}.sol:YourCollectible`
     } else {
-      it("Should deploy YourCollectible", async function () {
-        const YourCollectible = await ethers.getContractFactory("YourCollectible");
-        myContract = await YourCollectible.deploy();
-      });
+      contractArtifact = "contracts/YourCollectible.sol:YourCollectible";
     }
```
technophile-04, line 29: I think the reason we did it this way nd not going the `let contractArtifact: string;` route is : 

If you run `CONTRACT_ADDRESS=0x2a32333141E12DF1883A15537683A4bda72b9D2c yarn test --network sepolia`

This code will connect to that address on Sepolia (Instead of deploying a new copy of the contract) and then do other transactions on that deployed adderess, while in `let contractArtifact: string;` will deploy new contrat and do transaction on that new deployed instance. 

Lol both approches make sense to me but I think going `let contractArtifact: string;` (this PR changes) will make our life easy too

carletex:
Updated the top comment!

The ([updated](https://github.com/austintgriffith/speedrun-grader/pull/9/files)) autograder seems to be working with this change:

image

---

Now we can create them for the other challenges.

It'd also be nice if some of you ran the autograder locally & test... so you get some «knowledge» on it.

Thank you!

rin-st:
> ### So conclusion :
> 1. Remove top comment from test and merge this PR
> 2. Remove top comment and follow `let contractArtifcat: string` logic for other challenges 🙌

Agree with you guys!

> It'd also be nice if some of you ran the autograder locally & test... so you get some «knowledge» on it.

Thanks for the link to autograder, now I know how that magic works!  I've read the code and tested it, and your changes working great for challenge 0 👍 

great job!

technophile-04 (approved):
> Thanks for the link to autograder, now I know how that magic works! I've read the code and tested it, and your changes working great for challenge 0 👍

++, just tested autograder with these tests and works like a charm, Tysm 🙌

Merging this and will create issue for updating the same for other challenges and base-template

carletex:
Thank you both!!

We can also include it here: https://github.com/scaffold-eth/se-2-challenges/pull/136

PR #140, opened 2024-02-23
Ch2: Make it autograder ready

carletex:


rin-st (approved):
Tested and all tests passed! But I needed to find in Readme how YourTokenAutograder should work )
Lgtm!

PR #143, opened 2024-02-27
Chall5: Make it autograder ready

technophile-04:
### Description: 

Screenshot 2024-02-27 at 4 56 13 PM

Does not break anything both locally while running test and on autograder also works nicely 🙌

### DIFF
--- a/packages/hardhat/test/Challenge5.ts
+++ b/packages/hardhat/test/Challenge5.ts
@@ -1,13 +1,3 @@
-//
-// this script executes when you run 'yarn test'
-//
-// you can also test remote submissions like:
-// CONTRACT_ADDRESS=0x43Ab1FCd430C1f20270C2470f857f7a006117bbb yarn test --network rinkeby
-//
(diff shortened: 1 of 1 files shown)


carletex:
Wondering if we should change the ID (9 => 5)

We'd need to do it in the Autograder + the SRE repo.

It's an internal thing... but I might sleep better haha

What do you think?

technophile-04:
> Wondering if we should change the ID (9 => 5)
> We'd need to do it in the Autograder + the SRE repo.
> It's an internal thing... but I might sleep better haha
> What do you think?

+1 for change the ID (9 => 5) and +100 for Carlo's better sleep 😂, We should def do it if it doesn't have lot of side effects 🙌

carletex:
See

https://github.com/BuidlGuidl/SpeedRunEthereum/pull/198
https://github.com/austintgriffith/speedrun-grader/pull/9/commits/bb459dffcc2129b28988c0cfd4e0811bd1e40886

PR #146, opened 2024-03-03
Removed ownable from contract

miki-saarna:
Within `Checkpoint 2: ⚖️ Vendor 🤖`, the instruction `Edit Vendor.sol to inherit Ownable` assumes that the Vendor contract isn't already Ownable. Therefore, the Vender contract should not inherit the Ownable contract by default. Instead, the user should update the contract to inherit the Ownable contract when the instruction `Edit Vendor.sol to inherit Ownable` is reached.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ await yourToken.transfer(
 
 `contract Vendor is Ownable {`
 
-⚠️ You will also need to uncomment the Ownable.sol contract!
+⚠️ You will also need to uncomment the import of Ownable.sol contract!
 
 In `deploy/01_deploy_vendor.js` you will need to call `transferOwnership()` on the `Vendor` to make _your frontend address_ the `owner`:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -135,7 +135,7 @@ await yourToken.transfer(
 
 `contract Vendor is Ownable {`
 
-⚠️ You will also need to import the Ownable.sol contract!
+⚠️ You will also need to uncomment the Ownable.sol contract!
```
rin-st, line 138: Basically it's uncommenting of import, so probably `You will also need to uncomment the import of Ownable.sol contract!` is bettter?

rin-st:
Thanks for your pr @miki-saarna , nice catch! Please check my comment

miki-saarna (commented):
@rin-st great catch on making the minor adjustment within the documentation!

rin-st:
Updated uncomment message as mentioned here https://github.com/scaffold-eth/se-2-challenges/pull/146#discussion_r1510291060

Merging this, thank you again!

PR #147, opened 2024-03-03
Challenge 4 next14 migration

rin-st:


technophile-04 (approved):
Thansk @rin-st, just some minor missing parts which was missing but everything was working nicely !! 

Also tried speedrunning through challenge and everything works well 🙌

Here is my live deployed version : https://next14-chall-4.vercel.app/
And contract : https://sepolia.etherscan.io/address/0xDB5506d2Af060C4F92C277f37194EEC25915dBBF

Issue #157, opened 2024-03-11
Backmerge vercel deployment fix


technophile-04:
### Description : 

We recently merge https://github.com/scaffold-eth/scaffold-eth-2/pull/757, let backmerge it  in all challenges also remove #91 from challenge 0

technophile-04:
We should merge https://github.com/scaffold-eth/scaffold-eth-2/pull/780 instead

rin-st:
Completed!

PR #153, opened 2024-03-06
add vercel.json with mention installCommand

technophile-04:
### Description : 

Vercel now uses npm install by default and we get this : 

image

PR #158, opened 2024-03-21
feat: backmerge 21/03/2024

rin-st:
first step of https://github.com/scaffold-eth/se-2-challenges/issues/157


Pabl0cks (commented):
TYSM @rin-st for start working in the backmerge to the challenges!

I've been trying to reproduce the vercel deployment problem in the `base-challenge-template` but it's working both from CLI with `yarn vercel` and connecting the Github repo to Vercel. Backmerge branch is working good also in both cases.

Did you manage to reproduce the Vercel problems in the `base-challenge-template` branch? Maybe I'm doing something wrong 😥

rin-st:
> Did you manage to reproduce the Vercel problems in the base-challenge-template branch? Maybe I'm doing something wrong 😥

I didn't even test `base-challenge-template`, because
- as I remember problem appears only for prod branch  - `main` in most of the cases
- it seems https://github.com/scaffold-eth/scaffold-eth-2/pull/757/files which is caused the problem also in this pr
- if Shiv said that we need backmerge I believe he is right 💯 😄

Pabl0cks (approved):
> if Shiv said that we need backmerge I believe he is right 💯 😄

💯 I would even send him my private keys if he asked so  :)

Tested with a dummy page and everything looks good to me. `yarn test` and vercel deploys working well too.

technophile-04 (approved):
> if Shiv said that we need backmerge I believe he is right 💯 😄

> 💯 I would even send him my private keys if he asked so :)

🤣🤣🤣 🫶 lol, really sorry I didn't provide full context 

Actually when you deploy your app to vercel through CLI, vercel now uses `npm install` as default instead of `yarn install` to install dependencies. 

Although using `npm install` works in most of the cases (Like in this case bare se-2) sometimes when you have other dependencies it will break on vercel prod even though it works nicely locally (since locally we used yarn as package manger) an good example of this is https://github.com/scaffold-eth/se-2-challenges/pull/153

Hence in #153 and in [SE-2#780](https://github.com/scaffold-eth/scaffold-eth-2/pull/780/files#diff-457edfed186d364f8e7439adfa650514b59d8954769f7deb88240ca9359849b5) we explicitly tell vercel to use `yarn install` so that we are relaxed and more sure it will deploy properly since it worked locally.  


Tysm @rin-st for tackling it ! and @Pabl0cks for testing ! 🙌

PR #161, opened 2024-03-28
Ch1 backmerge 21/03/24

rin-st:


technophile-04 (approved):
Working great ! Tysm @rin-st ! 

Just pushed a couple of commits, I think we should update this on every challenge i.e wrap every challenges write calls inside try catch (because of https://github.com/scaffold-eth/scaffold-eth-2/pull/758/files) if we don't  we get Next  unhandled error 

Let's update this in every challenge backmerge PR 🙌

rin-st:
> Let's update this in every challenge backmerge PR 🙌

Created [wrapper](https://github.com/scaffold-eth/se-2-challenges/blob/base-challenge-template/packages/nextjs/utils/scaffold-eth/common.ts#L5) for that type of functions and added it to challenges 0-2. 

Tried to use `fn.name` instead `errorMessageFnDescription` but it doesn't work well in our case since it returns name of initial function. Example:  
```
f1 = () => {};
f1 = f2;
// f1.name = 'f1'; f2.name = 'f1';
```

If it's ok I'll add it to other challenges too

Pabl0cks (approved):
Tested and worked nicely in local and sepolia+vercel 

TYSM @rin-st !!

PR #163, opened 2024-03-28
Ch3 backmerge 21/03/24

rin-st:


Pabl0cks (commented):
When you roll the dice or do a rigged roll, the dice stays spinning forever.

I think it's related with the `wrapInTryCatch` not letting the error fire (we stopping the dice on error).

Pabl0cks (approved):
Now both rolls are working great! Approving 🙌

PR #165, opened 2024-03-28
Ch5 backmerge 21/03/24

rin-st:


Pabl0cks:
Working fine after wrapping in try/catch 🙌

PR #171, opened 2024-04-24
Challenge 0 wagmi 2 backmerge

rin-st:


Pabl0cks (approved):
TY @rin-st it's looking nice, GJ!

Deploy [link](https://challenge-6-multisig-93xtaetes-tokodevs-projects.vercel.app/)
Contract [link](https://sepolia.etherscan.io/address/0xdDFcc62C6194E9fb99af54660fb1a79b77e81b32)

Just a couple of issues to check (nitpicks in this Challenge, but more to check if we got some bug to fix in SE-2 wagmiV2):

- When clicking on "Transfers" tab we get this error: `Error: Contract not found`, but the information loads correctly.
- When you paste an address on the "Transfer To" field for the first time after loading the "My NFTs" tab, we get a small flickering. [Video here](https://www.loom.com/share/20ed32e9a638476b8f1c04199d973b62?sid=e594ad83-cad7-41e1-817e-f4b96d28ca3a)

rin-st:
> When clicking on "Transfers" tab we get this error: Error: Contract not found, but the information loads correctly.

You mean error in console, right? Added a quick fix for now. But looks like it's an at least 5-month old error, will add pr to se-2 soon

It's the way event history hook works - it shows error to the console even when it's caught by `catch (err)` block
> It's because of skeleton loading - input becomes wider, and hence card too. You can try to use ens names inside input, you'll see that card becomes wide. It worked that way already, I don't see the good way to fix it. We can fix width of the image so it will not widen on ens loading. So it will look like
image

instead of
image

rin-st:
created an issue for image width fix, I think it's better to fix it a bit later https://github.com/scaffold-eth/se-2-challenges/issues/172

rin-st:
[useScaffoldEventHistory update pr](https://github.com/scaffold-eth/scaffold-eth-2/pull/827) (could be merged later)

Pabl0cks:
> You mean error in console, right? Added a quick fix for now. But looks like it's an at least 5-month old error, will add pr to se-2 soon

Yeah! Just checked the SE-2 PR 🔥♥.  Since we are still with Ch. 0, should we wait to merge it, and then backmerge it to base & chal0 so is fixed on all challenges in this PR iteration?

> It's because of skeleton loading - input becomes wider, and hence card too.

> created an issue for image width fix, I think it's better to fix it a bit later https://github.com/scaffold-eth/se-2-challenges/issues/172

Ok! It was a very nitpicky thing, only commented it because it seemed to me that in an older version of Chal. 0 it was not happening, if not I wouldn't have raised it here probably 🙌

rin-st:
> Yeah! Just checked the SE-2 PR 🔥♥. Since we are still with Ch. 0, should we wait to merge it, and then backmerge it to base & chal0 so is fixed on all challenges in this PR iteration?

Yes, let's wait then

> Ok! It was a very nitpicky thing, only commented it because it seemed to me that in an older version of Chal. 0 it was not happening, if not I wouldn't have raised it here probably 🙌

It exists since start :)

rin-st:
> Yeah! Just checked the SE-2 PR 🔥♥. Since we are still with Ch. 0, should we wait to merge it, and then backmerge it to base & chal0 so is fixed on all challenges in this PR iteration?

Added events hook change and https://github.com/scaffold-eth/scaffold-eth-2/pull/825 . Working fine for me, hope for you too 😄  

I think I'll merge it again together with challenges 1 and 2

Pabl0cks:
> Added events hook change and https://github.com/scaffold-eth/scaffold-eth-2/pull/825 . Working fine for me, hope for you too 😄

Working well to me!

>I think I'll merge it again together with challenges 1 and 2

💯

rin-st:
Added last changes, tested. Working fine.

Pabl0cks:
I'm getting a strange render problem with the copy address button with the "View QR Code" option from RainbowKit.

If I view QR from a Metamask wallet, it's fine, but if I'm connected to burner wallet, copy icon renders super small:

image

Pabl0cks:
The challenge is still working fine to me. Should we change the spacing in `AddressQRCodeModal.tsx` from `space-x-4` to 1 or 0 to fix that kind of edge cases?

If we decide to tweak something, I'd probably just do it in se-2 and backmerge it in the next big backmerge to challenges, since is a really minor bug.

rin-st:
> I'd probably just do it in se-2 and backmerge it in the next big backmerge to challenges, since is a really minor bug.

You can create pr to se-2 and we'll add it to the current backmerge. Looks like we need to fix other bugs related to wagmi 2 in the current backmerge anyway

Pabl0cks:
> You can create pr to se-2 and we'll add it to the current backmerge. Looks like we need to fix other bugs related to wagmi 2 in the current backmerge anyway

Tried reproducing it in SE-2, it's not happening for the same address. The problem is a combination of Challenges font and some longer weight addresses due to their characters size. 

Should we change spacing a bit in SE-2 to fix that edge case (and other posible builds with SE-2 and big weight fonts), or should just leave it like that.

rin-st:
I think it's better to fix it in se-2

Pabl0cks:
Did last tests, looking good! 🙌

Pabl0cks:
While doing a fast check with the last burner commits on the challenges, realized of a [SE-2 issue with the burner](https://github.com/scaffold-eth/scaffold-eth-2/issues/843), a bit edge case, but maybe is good to add the fix when we have it.

rin-st:
Merging this since the burner issue is closed. Thank you a lot @Pabl0cks ! ❤️

PR #174, opened 2024-04-28
Challenge 1 wagmi 2 backmerge

rin-st:


Pabl0cks (commented):
Is looking good to me! The challenge is working fine, but found some edge case that we might want to check.

In `/staker-ui` if you change the `targetNetworks` from your connected network (hardhat for example), to another network (sepolia for example) you get this error instead of the "Wrong network" message from rainbow.

image

Doing the same at homepage, it works fine.

Pabl0cks (approved):
> Doing the same at homepage, it works fine.

TY! Working great now

rin-st:
- [ ] TODO: fix notification after update. Looks like we need to rewrite se-2 `getParsedError` function. But probably I'm wrong

Снимок экрана 2024-05-01 в 19 55 03

rin-st:
fixed by locking viem version

rin-st:
Added last changes, updated `useBalance` to `useWatchBalance`, tested. Working fine.

Pabl0cks:
Just did a final test and looking fine to me 🙌

PR #177, opened 2024-05-07
Challenge 4 wagmi 2 backmerge

rin-st:
Sorry for commits list, looks like I need to merge it without a squash commit and it will be better next time


Pabl0cks (approved):
Tested and working nicely to me! TYSM Rinat 🙏

PR #178, opened 2024-05-10
Challenge 5 wagmi 2 backmerge

rin-st:
note: currently I'm not using `burner-connector` (but added it to `package.json`) since in this challenge we need a new burner account in every tab. So I use mostly copypasted code with sessionStorage instead of localStorage. Hopefully we'll add a possibility to use sessionStorage in next versions of `burner-connector`.

note 2: sorry again for commits list, it looks I need to merge pr without squash


Pabl0cks (commented):
TYSM Rinat!! Been playing with the State Channels, all mechanics seem to work OK, but after you play for a bit, some weird behaviors appear with the channels. Some examples:

- First case. I got 3 rubes, everything working fine. I challenge 2nd one, and the 3rd rube starts seeing the wisdom from the 2nd rube instead of his. In the Guru I see the 3 rube wisdoms OK.

- Second case. I got 3 new rubes, everything starts OK, then I turn autopay off on the 1st one, and the 3rd starts seeing the 1st wisdom instead of his one
- I keep playing a bit more, click "Challenge" on the 2nd Rube, and that Rube dissapears from Guru interface
- After you do a `/yarn deploy --reset` the automatic refresh when you open a channel, stops working. You only see the channel opened when you manually refresh the browser. Before resetting the contract it would load OK after clicking the button. In the Guru UI they appear OK.

rin-st:
Great findings @Pabl0cks ! 

I refactored `Streamer` page and divided it into `Guru` and `Rube` components as I wanted for months 😄 . Now it's much easier to understand the logic. I found a bug which caused errors and hopefully 1-3 are fixed. Regarding 4 it didn't work from the start. I think we need to create an issue for it?

Also I changed `Readme.md` (see last commit). If you think wording is not very good please feel free to update it. 
Thanks!

packages/nextjs/components/streamer/Guru.tsx
```diff
@@ -0,0 +1,146 @@
+import { type FC, useEffect, useState } from "react";
+import { Address } from "../scaffold-eth";
+import { CashOutVoucherButton } from "./CashOutVoucherButton";
+import { Address as AddressType, formatEther, parseEther, verifyMessage } from "viem";
```
Pabl0cks, line 4: Did you not import `encodePacked`, `keccak256` and `arrayified` in purpose to not give too much clues to speedrunners to implement the Checkpoint 3, or just forgot it?

Pabl0cks:
> I refactored Streamer page and divided it into Guru and Rube components as I wanted for months 😄 . Now it's much easier to understand the logic. I found a bug which caused errors and hopefully 1-3 are fixed. Regarding 4 it didn't work from the start. I think we need to create an issue for it?

> Also I changed Readme.md (see last commit). If you think wording is not very good please feel free to update it.

Awesome job! Now is working like a charm, even the 4th point is working fine to me now 🙌 

Just added a small comment to check.

packages/nextjs/components/streamer/Guru.tsx
rin-st: Thanks! Updated [67ed995](https://github.com/scaffold-eth/se-2-challenges/pull/178/commits/67ed9953c3c30da67ffb6dbb369da426cc30b5b3)

Pabl0cks (approved):
Changes and new Readme looks good to me,

Just pushed a small change on Readme to delete the hint to uncomment `defundChannel` since it doesn't apply now with Wagmi v2, and we already hinted to uncomment the closure button before.

================================

I'm wondering if on the App Router migration, we should have moved the page specific components to a `_components` folder into the route, like we do with `blockexplorer` and `debug`, in this case would be for `streamer`, but probably other challenges need tweaks if we decide to implement the change.

rin-st:
Created an issue for it, will fix it a bit later https://github.com/scaffold-eth/se-2-challenges/issues/180

PR #179, opened 2024-05-11
Challenge 6 wagmi 2 backmerge

rin-st:


Pabl0cks (commented):
It's looking good to me, just a couple of issues to check:

- When you "Exec" a transaction from the pool, it disables the "Exec" button (OK), but it does not auto refresh the status to "Completed" (and other posible transactions with the same nonce to "Outdated"). You need to refresh the page to see the status changes.
- When you go to "Create" tab, it is perma flickering changing the "Select method" dropdown. I think the first time I opened the tab it was working OK (not 100% sure), but can't reproduce when it started happening, now it's happening all the time even refreshing the contract, deployment and backend.
<details><summary> Video of flickering </summary>

create-tab-infinite-flickering
</defails>

rin-st:
lol found a bug in a first frame of the video - `undefined` in input (it's very fast in real app so it hard to notice). Fixed it too

Pabl0cks:
> lol found a bug in a first frame of the video - undefined in input (it's very fast in real app so it hard to notice). Fixed it too

Good catch! I did not realize it, god bless low framerate :) 

Everything working fine now! GJ

PR #182, opened 2024-05-31
Fix incorrect component path in README

cleanerzkp:


Pabl0cks (approved):
Good catch @cleanerzkp !! 

Pushed a couple more links that had the old paths. Merging this! 👌

Pabl0cks:
> missed those, good update

No worries, we missed ALL of them previously ;) TYSM!

PR #194, opened 2024-08-13
Challenge 3 (Dice Game): Fix address-related errors/warnings

0xNeshi:
- **Mark `args.player` as `viem/Address`** - fixes this error:
image

- **Convert `Winner.address` type from `string` to `viem/Address`** - more accurate
- **Mark `args.winner` as `viem/Address`** - similar to first point


rin-st:
Updated one import and now it works as expected. Thank you @0xNeshi !

PR #199, opened 2024-08-14
Minor refactor to MetaMultiSigWallet.sol

0xNeshi:
- use `call` instead of `transfer` for sending ETH
- format code


rin-st (approved):
Thanks @0xNeshi ! Lgtm!

PR #200, opened 2024-08-15
fix: update deployment script names to match latest

MexicanAce:
# What :computer: 
* Update mentions of `01_deploy_vendor.js` to `01_deploy_vendor.ts`

# Why :hand:
* These scripts are now TypeScript files and this provides more clear instructions


rin-st:
Thank you @MexicanAce !

PR #201, opened 2024-09-03
fix: contract method

matteo-brandolino:
I have fixed a typo in the Challenge 1 test file by replacing mintItem with stake


technophile-04 (commented):
Nice catch! Thanks @matteo-brandolino!

PR #203, opened 2024-09-13
Update vercel

piersy:
Updates vercel to support git submodules and worktrees, prior to this update 'yarn vercel' would not work if the repo was checked out as a submodule or worktree.


egrojleinad (commented):
ok

rin-st:
We're moving challenges to `create-eth` extensions so vercel and other deps versions will be up to date with `create-eth`. So closing this, but thank you for PR @piersy !

PR #218, opened 2024-10-16
Challenge 1: Update readme for op sepolia

rin-st:


Pabl0cks (approved):
LGTM! Only pushed a small typo in the block explorer URL 🙏

rin-st:
Good catch, thanks!

PR #225, opened 2024-11-15
Update hardhat.config.ts

letmejustputthishere:
current rpc provider for sepolia does not work


technophile-04:
Hey thanks @letmejustputthishere! We just set some alchemy defaults back again so that it works nicely now out of the box check #227

PR #227, opened 2024-11-20
Alchmey default

technophile-04:
### Description: 

Reference https://github.com/scaffold-eth/scaffold-eth-2/pull/987

PR #229, opened 2024-11-23
Add Troubleshooting Tip for yarn account Showing 0 Balance in challenge-0-simple-nft

Belloabraham121:
This PR adds a troubleshooting note to the documentation for challenge-0-simple-nft. It addresses the issue of yarn account returning a balance of 0 even when the wallet has Sepolia ETH. The fix involves correctly configuring the RPC provider URL by replacing it with a custom Alchemy API URL and ensuring the .env file contains the necessary keys (ALCHEMY_API_KEY and DEPLOYER_PRIVATE_KEY).

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,8 @@ yarn start
 
 > 🌉 If you need to bridge funds from Sepolia to OP Sepolia, you can use the [Superchain Bridges](https://app.optimism.io/bridge/deposit). This allows you to transfer ETH and other supported tokens between the two networks.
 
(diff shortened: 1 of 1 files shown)


carletex:
Hey @Belloabraham121 thanks for this!

A couple of months ago we removed the default alchemy key (because we were hitting some limits), so we had this issue with the Sepolia RPC not working.

But we added it back last week https://github.com/scaffold-eth/se-2-challenges/pull/227, so this shouldn't be happening.

Thanks again!

Issue #234, opened 2024-11-27
Migration to extensions checklist


rin-st:
Use this as example [Challenge 0 example](https://github.com/scaffold-eth/create-eth-extensions/pull/29)

Checklist
**1. Branches**
Use

```
git switch --orphan challenge-[n]-[challenge-name]--extension
```

To create the branch, so the new branch has no files or history (we don't need a history in this branch because we won't merge this branch to main).

Add an empty commit and push the branch:

```
git commit --allow-empty -m "Init challenge n extension"
git push
```

And then use this branch as the starting branch to build the extension running:

```
git checkout -b chn-extension-init
```

From challenge-[n]-[challenge-name]--extension branch.

- [ ] Completed

**2. README.md**
Surface structure of extension:
```
├── README.md
└── extension
    ├── README.md.args.mjs
    └── packages
```

Copy `README.md` from current challenge version and place it in the root of extension. This README will be used by the SRE website.

Change this part:
```
git clone https://github.com/scaffold-eth/se-2-challenges.git challenge-0-simple-nft
cd challenge-0-simple-nft
git checkout challenge-0-simple-nft
yarn install
```
to:
```
npx create-eth@latest -e sre-challenge-0 challenge-0-simple-nft
cd challenge-0-simple-nft
```
- [ ] Completed

**3. README.md.args.mjs**
Create `README.md.args.mjs` inside `extension` folder. Structure of the file:
```
export const skipQuickStart = true;

export const extraContents = `... `;
```
We're setting `skipQuickStart` to `true` since the default quick start is not very good for us, as in the case of extensions it will be written before the description.

`extraContents` is mostly the same as main `README.md`, except `## Checkpoint 0: 📦 Environment 📚`: node/yarn/git and npx create-eth parts are removed since we don't need those parts from inside an already created instance. Also `yarn chain/deploy/start` is a bit different. 
Use `\` before backticks

- [ ] Completed

**4. packages/hardhat**
Add your
- [ ] contracts
- [ ] deployments files
- [ ] tests
- [ ] (optional) (?) `openzeppelin` contracts mostly uses `^0.8.20`, so probably you need to change solidity versions in contracts. For example I needed to change version from 0.8.4 to 0.8.20 in challenge 2 since it uses `Ownable` of `openzeppelin`
- [ ] (optional) add deps to package.json if needed

**5. packages/nextjs**
Add your
- [ ] pages/components/api routes
[...]

PR #250, opened 2024-12-16
Challenge 5 migration to extension

### LINKED ISSUE #246: Migrate challenge 5 to extension

damianmarti:
Challenge 5 migrated to an extension following https://github.com/scaffold-eth/se-2-challenges/issues/234

Install from create-eth repo with:

```
yarn cli -e scaffold-eth/se-2-challenges:ch5-extension-init
```

closes #246


rin-st:
Houston, we have a problem

https://github.com/scaffold-eth/se-2-challenges/blob/challenge-5-state-channels/packages/nextjs/services/web3/wagmiConnectors.tsx#L1

In this challenge we don't use `burner-wallet` package, I used sessionStorage version instead of localStorage used in `burner-wallet`. I did it because we needed to get a new user in every browser tab.

So, options we have
- (most easy) ask to open Rubes in a new private browser windows
- update `burner-wallet` and add possibility to use sessionStorage
- try to use the same sessionStorage version
- ... something different

extension/README.md.args.mjs
```diff
@@ -0,0 +1,275 @@
...
+- ⏱ Create a Challenge mechanism with a timeout, so that **rubes** are protected from a **Guru** who goes offline while funds are locked on-chain (either by accident, or as a theft attempt).
+- ⁉ Consider some security / usability holes in the current design.
+
+> 💬 Meet other builders working on this challenge and get help in the [State Channel Telegram](https://t.me/+k0eUYngV2H0zYWUx)!
+
+---
+
+## Checkpoint 0: 📦 Environment 📚
+
+> in the same terminal, start your local network (a blockchain emulator in your computer):
```
rin-st, line 32: remove "In the same terminal" part

damianmarti:
> Houston, we have a problem
> 
> https://github.com/scaffold-eth/se-2-challenges/blob/challenge-5-state-channels/packages/nextjs/services/web3/wagmiConnectors.tsx#L1
> 
> In this challenge we don't use `burner-wallet` package, I used sessionStorage version instead of localStorage used in `burner-wallet`. I did it because we needed to get a new user in every browser tab.
> 
> So, options we have
> 
> * (most easy) ask to open Rubes in a new private browser windows
> * update `burner-wallet` and add possibility to use sessionStorage
> * try to use the same sessionStorage version
> * ... something different

oh! 

I think that we can go with the easier solution, asking to open the other user in a private browser window...

extension/README.md.args.mjs
damianmarti: Fixed!

Yes, it is the same as in some of the previous extensions I will fix it there too!

rin-st:
> I think that we can go with the easier solution, asking to open the other user in a private browser window...

Yes, let's try to do it that way. Update Readme with that info please. Keep in mind we need a new private browser window for every rube, different tabs of one window also doesn't work

damianmarti:
> > I think that we can go with the easier solution, asking to open the other user in a private browser window...
> 
> Yes, let's try to do it that way. Update Readme with that info please. Keep in mind we need a new private browser window for every rube, different tabs of one window also doesn't work

Checking the README I read:

⚠️ Note: previous challenges created new addresses by opening an incognito window or a different browser. This will not work for this challenge, because the off-chain application uses a very simple communication pipe that doesn't work between different browsers or private windows.

What is the limitation to making it work on different browsers or private windows?

rin-st:
Oh, it uses https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API (grep `BroadcastChannel` in the app).

As I remember it worked only with tabs before, but now I see in the docs:

"The Broadcast Channel API allows basic communication between [browsing contexts](https://developer.mozilla.org/en-US/docs/Glossary/Browsing_context) (that is, **windows**, tabs, frames, or iframes) and workers on the same [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin)."

So probably windows also works. Will try it tomorrow

rin-st:
Tried fast check, doesn't work (. Will play with it tomorrow

Also, this challenge uses `ethers`, `humanize-duration` and `@types/humanize-duration` deps. Uncomment checkpoint 4 and you'll see

damianmarti:
> Tried fast check, doesn't work (. Will play with it tomorrow

I tried the broadcast channel with the incognito window and it's not working :-(

We will need to go with one of the other options:

1. update burner-wallet and add the possibility to use sessionStorage
2. try to use the same sessionStorage version
3. ... something different

Maybe we can try option 1. 

Another option is to remove the broadcast channel and use a backend for the messages. 

But adding the sessionStorage to the burner wallet may be something useful for other projects too.

@rin-st what do you think?

rin-st:
> Maybe we can try option 1.

Yes, I'll try to update it later today

damianmarti:
> > Maybe we can try option 1.
> 
> Yes, I'll try to update it later today

Great! Thanks!

I will try to remove the ethers dependency, it seems like the missing function from viem is already implemented https://viem.sh/docs/utilities/parseSignature

damianmarti:
Looking at the burner wallet code from the previous challenge 5

> > > Maybe we can try option 1.
> > 
> > 
> > Yes, I'll try to update it later today
> 
> Great! Thanks!
> 
> I will try to remove the ethers dependency, it seems like the missing function from viem is already implemented https://viem.sh/docs/utilities/parseSignature

Removed viem dependency using hexToSignature (we should update it to use parseSignature when updating viem).

Added humanize-duration dependency too and fixed some minor code comments.

rin-st:
> Removed viem dependency using hexToSignature (we should update it to use parseSignature when updating viem).

Great! But why not to use `parseSignature` right away? Our `viem` version allows it

We updated `burner-connector` to next version so it can work with sessionStorage. I tried it with this challenge and now it works as expected. But to make it installed by default we need to wait for https://github.com/scaffold-eth/scaffold-eth-2/pull/1021 to be merged. Also, after that merge, we need to add this to this challenge READMESs to make it work

---
add this line to `wagmiConnectors.ts` to use sessionStroage version and test: 

```ts
rainbowkitBurnerWallet.useSessionStorage = true;
```

damianmarti:
> > Removed viem dependency using hexToSignature (we should update it to use parseSignature when updating viem).
> 
> Great! But why not to use `parseSignature` right away? Our `viem` version allows it

I think the viem version we are using does not have parseSignature yet. I tried it before changing to hexToSignature and the method does not exist.

> 
> We updated `burner-connector` to next version so it can work with sessionStorage. I tried it with this challenge and now it works as expected. But to make it installed by default we need to wait for [scaffold-eth/scaffold-eth-2#1021](https://github.com/scaffold-eth/scaffold-eth-2/pull/1021) to be merged. 

Great!! Thanks!!

Also, after that merge, we need to add this to this challenge READMESs to make it work
> 
> add this line to `wagmiConnectors.ts` to use sessionStroage version and test:
> 
> ```ts
> rainbowkitBurnerWallet.useSessionStorage = true;
> ```

I will add this line to the READMEs!

damianmarti:
@rin-st Added the rainbowkitBurnerWallet config to READMEs!

The PR you mentioned was merged, but now we have to wait to get this update on create-eth, right?

rin-st:
> I think the viem version we are using does not have parseSignature yet. I tried it before changing to hexToSignature and the method does not exist.

It works for me :shrug: . Added it [bb5cdf9](https://github.com/scaffold-eth/se-2-challenges/pull/250/commits/bb5cdf985ac7dbebb51ac2284f9225cba5218ba7), please try on your side 

> The PR you mentioned was merged, but now we have to wait to get this update on create-eth, right?

Yes, sorry :)

damianmarti:
> > I think the viem version we are using does not have parseSignature yet. I tried it before changing to hexToSignature and the method does not exist.
> 
> It works for me 🤷 . Added it [bb5cdf9](https://github.com/scaffold-eth/se-2-challenges/pull/250/commits/bb5cdf985ac7dbebb51ac2284f9225cba5218ba7), please try on your side

Working fine! Thanks!!

damianmarti:
> 
> > The PR you mentioned was merged, but now we have to wait to get this update on create-eth, right?
> 
> Yes, sorry :)

Create-eth with the new burner connector was released and working great with the challenge! cc/ @rin-st @Pabl0cks

extension/packages/nextjs/app/streamer/page.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const { data: closedHistoryData } = useScaffoldEventHistory({
+    contractName: "Streamer",
+    eventName: "Closed",
+    fromBlock: 0n,
+    watch: true,
+  });
+
+  useEffect(() => {
+    if (closedHistoryData?.length !== undefined && closedHistoryData?.length !== closed.length) {
+      const closedChannelsAddresses = closedHistoryData?.map(event => event.args[0]);
```
Pabl0cks, line 56: Same:

```
const closedChannelsAddresses = closedHistoryData?.map(event => event.args[0]) as AddressType[];
```

extension/packages/nextjs/app/streamer/page.tsx
```diff
@@ -0,0 +1,97 @@
...
+    eventName: "Opened",
+    fromBlock: 0n,
+    watch: true,
+  });
+
+  const [opened, setOpened] = useState<AddressType[]>([]);
+
+  useEffect(() => {
+    if (openedHistoryData?.length !== undefined && openedHistoryData?.length !== opened.length) {
+      const openedChannelsAddresses = openedHistoryData?.map(event => event.args[0]).reverse();
```
Pabl0cks, line 73: Same:

```
const openedChannelsAddresses = openedHistoryData?.map(event => event.args[0]).reverse() as AddressType[];
```

extension/README.md.args.mjs
Pabl0cks: Can you check if the update is pushed?

Pabl0cks (commented):
Nice job @damianmarti and @rin-st !! 

With latest create-eth update and the new `rainbowkitBurnerWallet.useSessionStorage = true;` Readme hint is working great to me. 

Only added a couple nitpick comments.

rin-st (approved):
Lgtm! Just update types Pablo mentioned, please 🙂

extension/README.md.args.mjs
damianmarti: Sorry, I don't know what happened. Pushed now.

extension/packages/nextjs/app/streamer/page.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const { data: challengedHistoryData } = useScaffoldEventHistory({
+    contractName: "Streamer",
+    eventName: "Challenged",
+    fromBlock: 0n,
+    watch: true,
+  });
+
+  useEffect(() => {
+    if (challengedHistoryData?.length !== undefined && challengedHistoryData.length !== challenged.length) {
+      const challengedChannelsAddresses = challengedHistoryData?.map(event => event.args[0]);
```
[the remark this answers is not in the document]
damianmarti: Did you get these errors when deploying to Vercel? I wasn't getting these errors, but I added the type anyway, it makes sense.

Thanks!!

extension/packages/nextjs/app/streamer/page.tsx
damianmarti: Added type.

extension/packages/nextjs/app/streamer/page.tsx
damianmarti: Added type.

damianmarti:
> Nice job @damianmarti and @rin-st !!
> 
> With latest create-eth update and the new `rainbowkitBurnerWallet.useSessionStorage = true;` Readme hint is working great to me.
> 
> Only added a couple nitpick comments.

Thanks!!

Fixed! Please take a final review when you have some time. Thanks!

Pabl0cks (approved):
LGTM! Thx Damu 🙏

extension/packages/nextjs/app/streamer/page.tsx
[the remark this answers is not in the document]
Pabl0cks: Yep! Vercel was failing to me

PR #260, opened 2025-01-16
Update readme to add checks for balance and allowance

### LINKED ISSUE #259: Challenge 4. Generic error when there is not enough allowance to transfer tokens
The old `DEX.sol` solutions may not throw the right errors now, because of the new error handling in OpenZeppelin  5.0.  See https://github.com/scaffold-eth/se-2-challenges/issues/247

We need to check for token allowance and throw an specific error, otherwise we may end up in something like:

Image

We could add some hint in Challenge 4 readme about that.

Pabl0cks:
Add extra questions and updated solution snippets to check user balance and contract allowance before executing the transactions

Fixes #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,7 +375,7 @@ The basic overview for `ethToToken()` is we're going to define our variables to 
 
 <details markdown='1'><summary>Question One</summary>
 
-> How would we make sure the value being swapped for ETH is greater than 0?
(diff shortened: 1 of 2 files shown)


rin-st (approved):
Lgtm! Thank you!

damianmarti (approved):
Tested the challenge with the new contract code added to the README and I'm getting the right error messages now!

Thanks @Pabl0cks !!

PR #280, opened 2025-03-15
Over-collaterlized Lending Challenge

escottalexander:
This PR can be tested with `npx create-eth -e escottalexander/se-2-challenges:2-over-collateralized-lending`.

Open to any feedback you may have. The only major difference style-wise (that I am cognizant of) is the omission of the hero.png banner which Carlos and I discussed as being unnecessary.


escottalexander:
Initial feedback from Azimov:
"Here are my comments. I hope they are useful:

1. Screenshots and my frontend are different, it says:  Below is what your front-end will look like with no implementation code ... (there are differences between the current app and screenshot, maybe worth to update)

2. Minor thing, in the frontend, it is hard to distinguish the text "Amount" from the background. Also, the same for the health factor (when it is green and a bit when orange). Also, in the table for orange and green.

3. I didn't find the solution for chekpoint 2, is it okay? I can see solutions for other challenges

3. Side quest 1. In Readme, it says 
Before we implement the logic we need to crete a new interface called IFlashLoanRecipient. You can define it beneath the Lending.
It was already implemented.

4. The rest is cool and very nice! I am very happy that it will published be soon. Thank you very much!"

escottalexander:
Just a note. I added the hero banner image to the readme but I don't think the url will work until this is merged.

rin-st:
Hey @escottalexander , we need this challenge to be also branched from `base-challenge-template`, not from `main`. So you can recreate both PR branches from `base-challenge-template` or change git history some way. Please lmk if you need help with it!

PR #284, opened 2025-04-24
only set contract state when on hardhat chain

escottalexander:
This just adds a check to see if the deployment is local so that we can avoid running the `hardhat_setBalance` methods when deploying to a real network. Ran into this yesterday while testing. I am inclined just to merge it since it is simple but I will let you guys check my logic as well.

Without this it still deploys fine but it gives a nasty error when it tries to use the `hardhat_setBalance` method on a real RPC.


rin-st (approved):
Lgtm!

PR #289, opened 2025-05-01
Filters liquidators to include only those who have enough corn balance

NikolaiL:
When running simulation experienced many liquidation errors. The reason is that everyone is included in the eligibleLiquidators list without checking CORN balance required for liquidation.

Added additional check for eligibleLiquidators to include only ones who have enough CORN to liquidate.

### DIFF
--- a/extension/packages/hardhat/scripts/marketSimulator.ts
+++ b/extension/packages/hardhat/scripts/marketSimulator.ts
@@ -194,8 +194,20 @@ async function checkAndPerformLiquidations(lending: Lending, corn: Corn, account
     const isLiquidatable = await lending.isLiquidatable(user);
     if (!isLiquidatable) continue;
 
-    const eligibleLiquidators = accounts.filter(account => account.wallet.address.toLowerCase() !== user.toLowerCase());
+    const liquidators = accounts.filter(account => account.wallet.address.toLowerCase() !== user.toLowerCase());
+    
+    const liquidatorsWithBalance = await Promise.all(
+      liquidators.map(async account => {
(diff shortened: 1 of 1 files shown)


extension/packages/hardhat/scripts/marketSimulator.ts
```diff
@@ -194,8 +194,20 @@ async function checkAndPerformLiquidations(lending: Lending, corn: Corn, account
...
     if (!isLiquidatable) continue;
 
-    const eligibleLiquidators = accounts.filter(account => account.wallet.address.toLowerCase() !== user.toLowerCase());
-
+    const liquidators = accounts.filter(account => account.wallet.address.toLowerCase() !== user.toLowerCase());
+    
+    const liquidatorsWithBalance = await Promise.all(
+      liquidators.map(async account => {
+        const cornBalance = await corn.balanceOf(account.wallet.address);
+        //console.log(`Account ${account.wallet.address} has ${ethers.formatEther(cornBalance)} CORN. required: ${ethers.formatEther(amountBorrowed)}`);
```
escottalexander, line 202: Can you remove this comment?

escottalexander (changes requested):
Thank for submitting this. It looks good. One small nit pick about removing the comment. Other than that this has been tested and works great!

extension/packages/hardhat/scripts/marketSimulator.ts
NikolaiL: removed as requested.

escottalexander (approved):
Perfect.

PR #297, opened 2025-06-12
Downgrade create-eth version and update images

escottalexander:
This PR adds the new images that will differentiate more from the Stablecoin challenge.

We also downgraded the create-eth version because the tailwind styles are not applied the same way in the latest versions.
Rinat has said we will update all the challenges in the future to work with later versions of create-eth. Small side effect is we can't use the challenge shortcut `-e challenge-NAME` until we are updated to use the later versions.


Pabl0cks (commented):
Been playing with the extension and I was facing the `indexedDB` error on logs, and also some hydration errors when refreshing homepage. I think we may receive some user reports if they face this errors, even if the dapp works.

This error got fixed in `0.2.2`, maybe we could force this version, adding the necessary tweaks to `globals.css` to use the correct styles from the old tailwind config? 

In my local deployment it seems to work nicely, not sure if old `create-eth` support that kind of changes in `globals.css` or we need to use the beta for it?

carletex:
Great stuff! Let's do a couple of rounds of testing with 0.2.3 (when released). In the meantime, I'll be preparing the SRE side of things

Thanks all!

Pabl0cks (commented):
After some testing here is what I've found:

- Getting these indexedDB ReferenceErrors when you do `yarn start`, then it doesn't show anymore => We can try to fix in future versions, not annoying since is only showing in the console and not being spammed.
- Space Grotesk font is not showing properly with the new tailwind version
- I'm getting this random error when I open the Dashboard tab, or refresh it, in both 0.1.8 and 0.2.3 (very random, haven't found a pattern)

image

rin-st:
Added small fix for table colors
<details><summary>Screenshots</summary>
<p>

Before:
Screenshot 2025-06-13 at 17 14 57
Screenshot 2025-06-13 at 17 14 52

After:
Screenshot 2025-06-13 at 17 14 14
Screenshot 2025-06-13 at 17 14 05


</p>
</details>

PR #301, opened 2025-06-19
Challenge prediction markets

phipsae:
Change css file to be compatible with new create-eth version 2.4.


Pabl0cks (approved):
LGTM! 🙌 

There are some minor styles that should be reviewed like some fonts and the line below "Buy Yes" and "Sell Yes", but I think we can address them in a future PR

image

PR #302, opened 2025-06-19
fix fonts

technophile-04:
### Description: 

fixed fonts we need to include them use Next Js way, with new daisy and tailwind config font importing directly in CSS file doesn't seem to work :


Pabl0cks (approved):
Merging 🙏 tysm Shiv!

PR #311, opened 2025-06-20
[Beta migration] challenge 4 dex

rin-st:


Pabl0cks (commented):
GJ Rinat! Works nicely to me, just 1 nitpick, not sure if it's going to be easy to reproduce:

- After interacting with the DEX/Events tabs for a bit, when I click on Debug Contracts, I get the message: `Address "" is invalid.`
- If I refresh the website with Ctrl+F5, the message disappears, until I interact with the DEX/Events tabs again.

rin-st:
oh man, it was difficult to debug, fixed [0c30a83](https://github.com/scaffold-eth/se-2-challenges/pull/311/commits/0c30a8341d4d5d29685a183d62f3191c59d50454)

without this after visiting dex page and then going to debug some queries was fetched even with `query.enabled: false`. It seems because we have similar queries in dex and debug read contracts page, and that `query.enabled: false` was ignored.

PR #300, opened 2025-06-19
Over collateralized lending challenge updates

escottalexander:
This PR covers feedback we received from several places.

You can test it with:
```
npx create-eth@0.2.4 -e escottalexander/se-2-challenges:over-collateralized-lending challenge-over-collateralized-lending
```

This will require a reset of the test in the auto-grader which I have made an issue about.

For anyone who tests this here is the updated contract that is deployed on Sepolia: https://sepolia.etherscan.io/address/0x3BD8ec3816875bd4cEFaf977aF3F9F320e07F998#code

Closes #299

PR #317, opened 2025-06-30
Changes of OCL challenge which was reverted

rin-st:
Revert revert of #300 , so PR is the same


escottalexander (approved):
This looks good. I added a small commit to remove a commented out line of code in the deploy script. Thanks Rinat!

PR #335, opened 2025-07-24
Rename OverCollateralizedLending.ts to Lending.ts

carletex:
See https://github.com/austintgriffith/speedrun-grader/pull/24/files#r2226408574


rin-st:
Lgtm!

> Is this standard now where we should make the test name match the contract name or was there some other reason for this?

Yes, while working on grader v2 we decided to rename test files for consistency. Before, different challenges tests had different naming. Also, I think it's a good practice - it's easy to understand the tests for which contract is it.

Please don't merge this PR, we'll do it a bit later

rin-st (approved):

PR #351, opened 2025-08-13
Fix: update borrowing and flash loan logic in README

escottalexander:
Somebody recently reported some issues with this challenge. This is all related to some changes we pushed through after Austin had some feedback a few weeks ago. The side quests would not work with the new changes. 

In addition to this it seems a test was also not updated (I am surprised by this since I am sure we would have run them when we made the changes).

We need to update the autograder with the new test as well.

Here is a contract that matches what the README give the user: https://sepolia.etherscan.io/address/0xE543ac072299c4222dECa0BbdFFC8fd551f15649#code

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,9 +309,9 @@ For this challenge we will not focus on the Lending aspect as much as the other 
         }
         s_userBorrowed[msg.sender] += borrowAmount; // Update user's borrowed corn balance
         _validatePosition(msg.sender); // Validate user's position before borrowing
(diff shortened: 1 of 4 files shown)


rin-st (approved):
Thanks for the updates! Tested, working great

carletex:
Let me merge this later today, so I'll update the autograder

carletex:
Autograder updated! tested it with the contract that Elliott provided (which, interestingly, was also working with the previous test)

thanks guys!

PR #369, opened 2025-10-04
fix: additional condition to prevent overflow

megabyte0x:


rin-st (approved):
Thank you @megabyte0x !

PR #377, opened 2025-10-31
Challenge oracles

escottalexander:
This can be installed with `npx create-eth@2.0.4 -e escottalexander/se-2-challenges:challenge-oracles challenge-oracles`

Solutions to the contracts can be found here: https://github.com/BuidlGuidl/oracle-challenge/tree/main/packages/hardhat/contracts

After this is merged we need to do the following:
- [x] Add `challenge-oracles` shortcut to create-eth (Already done as of v2.0.4)
- [ ] Merge SRE Metadata PR (https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/353)
- [ ] We need to update the database for SRE perhaps just setting the challenge to active vs "coming soon" (Maybe more that I am unaware of)
- [x] Autograder PR to add tests for the Optimistic Oracle https://github.com/austintgriffith/speedrun-grader/pull/35

I think these last few steps will require @carletex. Thank you for the help. Let me know if you need anything from me.


phipsae:
@rin-st just a quick update, @escottalexander is working separately on that in here https://github.com/BuidlGuidl/oracle-challenge/pull/65

escottalexander:
@phipsae Please check the latest commit. It should satisfy all you feedback. Here are the design choices I landed on for the Staking Oracle updates:
- Use ORA for staking (No more ETH)
- Provide a phase after 2/3 of nodes have reported where the median is calculated, slashing is then based on the median
- Slashed nodes no longer are removed from the price so the price being published and "finality" are the same moment now

Thank you for the feedback.

escottalexander:
Once Phillip has reviewed the latest version ([here](https://github.com/BuidlGuidl/oracle-challenge/pull/66)) then I can attempt to use this with the latest create-eth and fix any issues that occur. Once that is fine then it will be ready for final reviews before inclusion into SRE 🥲.

escottalexander:
@rin-st This challenge is finally ready for your review. It is pretty big compared to most of the challenges but it gives people a good overview of different types of oracles and their tradeoffs.

README.md
```diff
@@ -1,6 +1,2056 @@
...
+![WhiteListOracle](https://github.com/user-attachments/assets/1fabc2d8-a1a4-4b0e-b00c-49f3fefec7fc)
+
+👊 Notice how the onlyOwner modifiers are commented out to allow you to have full control. Try manually changing the price of individual SimpleOracle contracts and adding new oracle nodes to see how the aggregated price changes:
+
+1. **Change Prices**: Use the frontend to modify individual oracle prices
+
+2. **Add New Nodes**: Create new SimpleOracle contracts through the whitelist oracle
+
+3. **Observe Aggregation**: Watch how the median price changes as you add/remove oracles
+
```
rin-st, line 454: Image

nit: on the frontend price counted only for ~20 seconds, not 24

README.md
```diff
@@ -1,6 +1,2056 @@
...
+- Because the oracle data is useful, anyone who wants to digest the data will be incentivized to run the `recordBucketMedian` function.
+- Slashing decisions compare a node's report against the **recorded median**.
+
+🎯 **Your Mission**: Complete the missing function implementations in the `StakingOracle.sol` contract. The contract skeleton is already provided with all the necessary structs, events, and modifiers but you need to fill in the logic.
+
+🔍 Open the `packages/hardhat/contracts/01_Staking/StakingOracle.sol` file to implement the staking oracle functionality.
+
+### ✏️ Tasks:
+
+1. **Implement `getCurrentBucketNumber()`**
```
rin-st, line 493: All the functions already implemented in the `StakingOracle` contract, you need to remove them

README.md
```diff
@@ -1,6 +1,2056 @@
...
+
+- 🔍 It should check that the assertion exists and hasn't been proposed yet. Otherwise revert with `AssertionNotFound` or `AssertionProposed`
+
+- ⏱️ It should validate the timing constraints - it has to be after `startTime` but before the `endTime` or else revert with `InvalidTime`
+
+- 💸 It should enforce the correct bond amount is provided or revert with `InvalidValue`
+
+- ✍️ It should update the assertion with the proposal
+
+- ⏳ It should set the `endTime` to `block.timestamp + MINIMUM_DISPUTE_WINDOW`
```
rin-st: here and below, constant name is `DISPUTE_WINDOW`

rin-st (commented):
Thank you for all the updates, looking great!

But I understood one thing: currently, user can submit only one contract as the solution of the challenge, so not clear what should we do?
1. If we want to allow submitting all 3 contracts it requires the changes on the sre website and grader, not sure it will be easy.
2. Submit only last contract (OptimisticOracle). We have some options here to all the tests pass
2.1. To have solutions of SimpleOracle and WhitelistOracle in the challenge so grader will download them 
2.2 Have different test files so grader will download and use only OptimisticOracle tests
2.3 Have all SimpleOracle and StakingOracle tests commented or disabled by default. User will need to uncomment/make them work during the challenge

Or maybe you could find better way?

For me, 2.2 > 2.3 > 2.1 > 1. What do you think? Could you try to implement 2.2?

cc @carletex

carletex:
> 2. Submit only last contract (OptimisticOracle).

I'm also inclined to any of these, if possible.

escottalexander:
@rin-st @carletex The challenge only requires that the user deploy and submit the last challenge (OptimisticOracle). I am missing what your 2.2 suggestion means. Do you mean that the grader will point to a separate repo that only has the OO tests? If so, that seems better to me than making the user uncomment files. Please help me understand your idea.

Another thought would be to a add check to the top of the other two test files that simply returns if being run with CONTRACT_ADDRESS env variable set. This way when being run by the auto-grader, they would be skipped. Could that work?

Thanks for the feedback. I will work on the feedback you left on other places.

README.md
escottalexander: This is a notoriously difficult issue. I have made some changes to hopefully help but I think we will have to be OK with "good enough" on this issue specifically. 

The problem has to do with the blocktime of the node getting out of sync with system time. This happens whenever  multiple transactions are automined without 1 second of time passing in system time. This causes the node time to escalate ahead of system time every time it happens. 

We tried three different methods (including some really elaborate ones that were just too clunky) and what we have here is the cleanest solution we could find. It includes disabling automine for contract deployments and setting the block interval to 1 second. Still some situations cannot be avoided.

README.md
escottalexander: Good catch. Removed in latest commit.

escottalexander:
@rin-st I have made the adjustments you requested including my idea for skipping the other two tests when running on the autograder. Let me know if that still wont work and we will figure it out. Thanks!

README.md
rin-st: Now it works better, thanks!

rin-st:
> The challenge only requires that the user deploy and submit the last challenge (OptimisticOracle). I am missing what your 2.2 suggestion means. Do you mean that the grader will point to a separate repo that only has the OO tests? If so, that seems better to me than making the user uncomment files. Please help me understand your idea.

Oh, it's already implemented as I wanted, OptimisticOracle tests in a separate file. Don't know why I didn't see it, maybe I checked a different repo, sorry.

So I think this is ready to be merged. And then we prepare/merge all of these in parallel

> After this is merged we need to do the following:

> * [x]  SRE should already have the image and description but will need the rest of the metadata defined in [challenges.ts](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/blob/main/packages/nextjs/utils/challenges.ts) ([Add metadata for Oracle challenge BuidlGuidl/SpeedRunEthereum-v2#353](https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/353))
> * [ ]  We need to update the database for SRE perhaps just setting the challenge to active vs "coming soon" (Maybe more that I am unaware of)
> * [ ]  Autograder PR to add tests for the Optimistic Oracle

Merging, thank you!

PR #385, opened 2026-01-07
Challenge token vendor - README fix

escottalexander:
Noticed I accidentally left this section in the readme. It shows twice after being installed with create-eth.
image


rin-st (approved):
Thanks!

PR #386, opened 2026-01-07
Challenge dice game refresh

escottalexander:
Updated this challenge to match the general feel of the other challenges. Removed some statements that were old/outdated. The front end is pretty much untouched except that I changed "SpeedRun" to "Speedrun" and made it work with scaffold-ui.

This will require that the autograder uses the new test file since checkpoints were introduced to match the other challenges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Before you begin, you need to install the following tools:
 Then download the challenge to your computer and install dependencies by running:
 
 ```sh
-npx create-eth@1.0.2 -e challenge-dice-game challenge-dice-game
+npx create-eth@2.0.4 -e challenge-dice-game challenge-dice-game
 cd challenge-dice-game
 ```
 
@@ -167,7 +167,7 @@ You have beaten the game, but where is your money? Since the RiggedRoll contract
 #### Configuration of Third-Party Services for Production-Grade Apps.
 
 By default, 🏗 Scaffold-ETH 2 provides predefined API keys for popular services such as Alchemy and Etherscan. This allows you to begin developing and testing your applications more easily, avoiding the need to register for these services.
-This is great to complete your **SpeedRunEthereum**.
+This is great to complete your **Speedrun Ethereum**.
 
 For production-grade applications, it's recommended to obtain your own API keys (to prevent rate limiting issues). You can configure these at:
 
@@ -183,7 +183,7 @@ For production-grade applications, it's recommended to obtain your own API keys 
 
 Run the `yarn verify --network your_network` command to verify your contracts on etherscan 🛰
 
-👉 Search this address on [Sepolia Etherscan](https://sepolia.etherscan.io/) (or [Optimism Sepolia Etherscan](https://sepolia-optimism.etherscan.io/) if you deployed to OP Sepolia) to get the URL you submit to 🏃‍♀️[SpeedRunEthereum.com](https://speedrunethereum.com).
+👉 Search this address on [Sepolia Etherscan](https://sepolia.etherscan.io/) (or [Optimism Sepolia Etherscan](https://sepolia-optimism.etherscan.io/) if you deployed to OP Sepolia) to get the URL you submit to 🏃‍♀️[SpeedrunEthereum.com](https://speedrunethereum.com).
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ yarn start
 
 ## Checkpoint 1: 🎲 Dice Game
 
-🔍 Inspect the code in the `DiceGame.sol` contract in `packages / hardhat / contracts`
+🔍 Inspect the code in the `DiceGame.sol` contract in `packages/hardhat/contracts`
 
 🔒 You will not be changing any code in the `DiceGame.sol` contract in this challenge. You will write your own contract to predict the outcome, then only roll the dice when it is favourable.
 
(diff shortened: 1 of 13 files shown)


extension/README.md.args.mjs
```diff
@@ -71,28 +69,99 @@ yarn start
...
+    if (roll > 5) revert NotWinningRoll(roll);
+
+    diceGame.rollTheDice{value: required}();
+}
+\`\`\`
+
+</details>
+</details>
+
+🚀 To deploy your RiggedRoll contract, uncomment the appropriate lines in the \`01_deploy_riggedRoll.ts\` file in \`packages/hardhat/deploy\`
```
rin-st: do we need to suggest to `yarn deploy --reset` here? Yes, we have it at the top of readme but not sure everyone will remember it

extension/README.md.args.mjs
```diff
@@ -71,28 +69,99 @@ yarn start
...
 
 💸 You will need to send some funds to your RiggedRoll contract before doing your first roll, you can use the Faucet button at the bottom left of the page.
 
+🐞 Go to the \`Debug Contracts\` tab and try the \`riggedRoll\` method inside the \`RiggedRoll\` contract.
+
 ❓ If you're struggling to get the exact same random number as the DiceGame contract, try adding some \`console.log()\` statements in both contracts to help you track the values. These messages will appear in the Hardhat node terminal.
```
rin-st: I think we don't need this after the solution

extension/README.md.args.mjs
```diff
@@ -71,28 +69,99 @@ yarn start
...
 
 ### ⚔️ Side Quest
 
 - [ ] Add a statement to require \`address(this).balance >= .002 ether\` in your riggedRoll function. This will help prevent calling the \`rollTheDice()\` function without enough value.
```
rin-st: This side quest is not actual too

extension/README.md.args.mjs
rin-st: remove this block

README.md
```diff
@@ -181,7 +287,7 @@ For production-grade applications, it's recommended to obtain your own API keys
 
 ## Checkpoint 6: 📜 Contract Verification
 
-Run the `yarn verify --network your_network` command to verify your contracts on etherscan 🛰
+Run the `yarn verify--network your_network` command to verify your contracts on etherscan 🛰
```
rin-st: typo here, `verify --network` with space

extension/README.md.args.mjs
```diff
@@ -71,28 +69,99 @@ yarn start
...
+    uint256 available = address(this).balance;
+    if (available < required) revert NotEnoughETH(required, available);
+
+    bytes32 prevHash = blockhash(block.number - 1);
+    uint256 nonce = diceGame.nonce();
+    bytes32 hash = keccak256(abi.encodePacked(prevHash, address(diceGame), nonce));
+    uint256 roll = uint256(hash) % 16;
+
+    // Only roll when we know we’ll win (matches the provided tests)
+    if (roll > 5) revert NotWinningRoll(roll);
```
rin-st, line 134: nit: On the frontend we're showing A, B, .. F if roll is 10, 11 .. 15
Image

But here in the contract we're showing the number. 
Image

Not sure if we need to change something, cursor suggests to add helper

```
// Helper function to convert roll number to hex character (0-9, A-F)
    function rollToHex(uint256 roll) internal pure returns (string memory) {
        if (roll < 10) {
            // Convert 0-9 to "0"-"9"
            return string(abi.encodePacked(bytes1(uint8(48 + roll))));
        } else if (roll <= 15) {
            // Convert 10-15 to "A"-"F"
            return string(abi.encodePacked(bytes1(uint8(65 + roll - 10))));
        }
        return "Invalid";
    }
```

and if not winning

```
// Only roll when we know we'll win (matches the provided tests)
        if (roll > 5) revert NotWinningRoll(rollToHex(roll));
```

If we decide to use the helper I think we need to add it by default.

What do you think, do we need it or better leave it as is?

rin-st (commented):
Thank you Elliott!

Added some comments

extension/README.md.args.mjs
escottalexander: I am moving to before the solution because I think this might be useful for someone that really wants to stretch themselves without hints.

extension/README.md.args.mjs
escottalexander: Moving this to before the solution. (Not as a side quest though)

extension/README.md.args.mjs
escottalexander: Personally I think it is fine as-is. I think people already need to follow that 10 = A, 11 = B etc. so it is fine that the error reports the corresponding number instead of the hex. If you think this is a trip up for some users perhaps we could add a short statement to the beginning of the README that explains the dice is hex based but the contract just uses integers.

escottalexander: Adding this line to the section where we tell them to implement `riggedRoll()`:

🎲 Keep in mind the dice on the frontend is using hexadecimal characters but in the contract we can get by with integers. A,B,C,D,E,F = 10,11,12,13,14,15

escottalexander:
@rin-st Thanks for the great feedback. Everything you mentioned has been resolved though feel free to push back if you don't like the direction with any of the changes.

rin-st (approved):
Thanks for the updates!

PR #389, opened 2026-01-15
Use an interface in Decider to avoid file import

escottalexander:
The hope is that this will fix the autograder.


rin-st (approved):
Thank you!

PR #396, opened 2026-02-10
Base challenge template - AI guided

rin-st:
## AI-Guided Interactive Learning System for SpeedRunEthereum Challenges

   An AI tutor system that turns SE-2 challenges into interactive, guided learning experiences. Instead of reading docs and copying code, users learn blockchain concepts
    through conversation with an AI that teaches, asks questions, validates understanding, and provides progressive hints — all inside their code editor.

   ### How It Works

   Users run `/start` in their AI editor (Claude Code, Cursor). The AI reads a `CHALLENGE.yaml` config file, initializes progress tracking, and begins guiding the user
   through a series of **checkpoints** — teaching concepts, asking questions, and assigning coding tasks.

   Progress is saved to `.challenge-ai/progress.json`, so users can close their editor and resume anytime.

   ### Two Checkpoint Types

   **1. Concept Checkpoints** (`unlocks` field)
   - AI presents teaching material (context)
   - Asks conceptual questions to verify understanding
   - Evaluates answers against expected concepts (≥70% match = correct)
   - On correct answer: reveals and auto-injects code into the contract (replaces TODO markers)
   - Progressive hint system: context reference → provided hint → multiple choice → re-teach

   **2. Code-Writing Checkpoints** (`task` field)
   - AI presents teaching material, then assigns a coding task
   - User writes code in specified file, types `check` to run tests
   - On test failure: AI reads the contract and gives specific suggestions
   - `hint` for progressive hints from the challenge config
   - `/skip` to have AI write the solution with explanation
   - On test pass: celebrates, updates progress, moves to next checkpoint

   ### Challenge Definition (`CHALLENGE.yaml`)

   Challenge creators define their entire curriculum in a single YAML file:

   ```yaml
   challenge:
     name: "Challenge Name"
     checkpoints:
       - id: "checkpoint-1"
         title: "Learn about mappings"
         context: "Teaching material..."
         questions: [{text: "What is a mapping?", concepts: ["key-value", "storage"]}]
         unlocks: {file: "contract.sol", todo_marker: "TODO-1", code: "mapping(address => uint)..."}

       - id: "checkpoint-2"
         title: "Write a withdraw function"
         context: "Teaching material..."
         task:
           description: "Implement withdraw..."
           file: "contract.sol"
           test_command: "yarn hardhat test..."
           hints: ["Think about msg.sender", "Use transfer()"]
           solution: {code: "function withdraw()...", explanation: "..."}
   ```

   Optional `setup.template` transforms a complete contract into a TODO-marked skeleton that gets progressively filled in as the user advances.

   ### Multi-Editor Support

   Shared instruction files live in `extension/.ai/` and are referenced by editor-specific configs:

   | Feature | Claude Code | Cursor |
   |---|---|---|
   | Start command | `.claude/skills/start/` | `.cursor/commands/start.md` |
   | Skip command | `.claude/skills/skip/` | `.cursor/commands/skip.md` |
   | Progress tracker | `.claude/agents/progress-tracker.md` | `.cursor/agents/progress-tracker.md` |

   AI autosuggestions/autocomplete are **disabled by default** (via `.vscode/settings.json`) so users write code themselves rather than accepting AI-generated
   completions. The AI tutor chat itself remains fully enabled.

   ### File Structure

   ```
   extension/
   ├── .ai/                                    # Shared AI instructions (editor-agnostic)
   │   ├── CHALLENGE.yaml.example              # Template for challenge creators
   │   ├── instructions/
   │   │   ├── start-content.md                # Main tutor behavior guide
   │   │   └── skip-content.md                 # Skip/solution handling
   │   └── agents/
   │       └── progress-tracker-content.md     # Progress file operations
   ├── .claude/                                # Claude Code wiring
   │   ├── skills/{start,skip}/SKILL.md
   │   ├── agents/progress-tracker.md
   │   └── settings.local.json.template.mjs
   ├── .cursor/                                # Cursor wiring
   │   ├── commands/{start,skip}.md
   │   └── agents/progress-tracker.md
   └── .vscode/settings.json                   # Disables AI autosuggestions
   ```

   ### Key Design Decisions

   - **Teach first, then ask/code**: Context is always presented before any questions or tasks
   - **Never say "wrong"**: Encouraging tone, celebrate every checkpoint completion
   - **Progressive hints**: Multiple levels of help before revealing answers
   - **Stateful**: Progress persists across sessions via progress-tracker agent
   - **Challenge-creator friendly**: One YAML file defines the entire curriculum
   - **Autosuggestions disabled, AI chat enabled**: Prevents AI from writing code for the user while keeping the tutor active

   ### TODO
   - Change `AGENTS.md` to `AGENTS.md.args.mjs` ([template](https://github.com/scaffold-eth/create-eth/blob/main/templates/base/AGENTS.md.template.mjs)) when migrating
   to latest create-eth, remove CLAUDE.md


rin-st:
Added some changes to this PR. 

Again, since this PR to `base-challenge-template`, changes will be added to the challenge branches when backmerging. So to create AI verion of the new challenge, we mostly need to add `CHALLENGE.yaml`, and then some minor challenge-related changes.

I created the prompt and used it to create tokenization #399 and crowdfunding #400. Branches in the prompt should be changed when related PRs will be merged

```
Create a new branch challenge-<name>-ai from challenge-<name>. Merge learn-with-ai-combined into it (resolve conflicts if any).

  Then add a CHALLENGE.yaml for this challenge that enables the AI-guided flow. Use the branch's README.md as the source for challenge content/steps. Use
  CHALLENGE.yaml.example from learn-with-ai-combined as the primary template, and reference existing CHALLENGE.yaml files from other AI challenge branches (e.g.
  challenge-tokenization-ai, challenge-ai-crowdfunding) for real-world examples.

  Requirements:
  - The AI-guided flow (via CHALLENGE.yaml) must work
  - The existing non-AI flow (manual README-based) must remain unchanged

  Create a plan first, then execute step by step.
```

---

Answering to Pablo's AI report from https://github.com/scaffold-eth/se-2-challenges/pull/397#pullrequestreview-3784538560

> The experience of going to Claude/Cursor still may have too much friction for "builders", having a web UI is on the roadmap?

Yes, it will be next step

> Antigravity does not support this yet, right?

I don't know, for now I tested it only with claude/cursor

> ### Bugs / Broken Stuff
> #### 1. `yarn install` not run before first test
> First "check" fails — `node_modules` doesn't exist. Should run `yarn install` during setup or warn the user beforehand.

it should run when installing extension with create-eth, so I think users won't meet this

> #### 2. Hardhat config missing Solidity 0.8.20 compiler
> Contract uses `0.8.20` but config only has `0.8.30`. First compile fails with `HH606`. Should be pre-configured in the repo.

fixed in #400 

> #### 3. `.cursorignore` blocks contract file access in Cursor
> Cursor can't read `CrowdFund.sol` — filtered by `.cursorignore`. Breaks the whole Cursor experience.

cursorignore is removed in this PR

> #### 4. `yarn generate` / `yarn deploy` need interactive input
> Both prompt for passwords — crashes when the AI tries to run them. Should tell users to run these in their own terminal.

Claude answer: yarn generate requires an interactive password prompt which can't be handled in this non-interactive shell. You'll need to run yarn generate in your own terminal so you can enter a password to encrypt the private key.

I think it's enough?

> ### UX Issues (Claude Code)
> #### 5. User doesn't know WHERE to write code
> The very first reaction was "Not clear to me where do I need to write what." For someone using Claude Code (terminal only), being told "open packages/hardhat/contracts/CrowdFund.sol" might not be enough. Maybe it should hint something?

[bdf12bb](https://github.com/scaffold-eth/se-2-challenges/pull/396/commits/bdf12bba5ebada41c472dc580dedc5015e8660e0)

> #### 6. "hint" not visible after long teaching blocks
> Welcome message mentions "hint" but it scrolls off-screen after the big teaching wall. The closing prompt only says `say "ready"` — should also mention "hint", especially on that first overwhelming checkpoint.

[3b99c74](https://github.com/scaffold-eth/se-2-challenges/pull/396/commits/3b99c74341d4713d128f0a2b03ffadaa414714b8)

> #### 7. Checkpoint 4 task description gives away the answer
> Task says "just call `contribute()` from inside `receive()`" — that's literally the solution. User called it out. Make the description vaguer and let them figure it out.

It's almost the same in the `README.md`, so I think it's fine. In #400 wording is a bit different but still it's clear that you need to do the same

> #### 8. progress.json prompt confuses beginners
> Users see a big red/green diff and "Do you want to overwrite progress.json?" with no context. Consider auto-allowing progress file writes.

Tried to implement that (without template file in extension and with it), but it doesn't work for now

> #### 9. Completion message says `yarn deploy` without a network
> Deploys to localhost. Should say `yarn deploy --network sepolia`. (The AI caught this in conversation when asked — it's a CHALLENGE.yaml template issue.)

Currenly it's suggesting to check README for deploy info, probably I need to change it

> #### 10. Suggested faucets require mainnet balance
> Sepoliafaucet.com and Alchemy faucets need mainnet ETH. Use Google's instead: [https://cloud.google.com/application/web3/faucet/ethereum/sepolia](https://cloud.google.com/application/web3/faucet/ethereum/sepolia)

Currenly it's suggesting to check README for deploy info, probably I need to change it


---

> ### UX Issues (Cursor). Probably most are caused because the `cursorignore`?

> #### 11. Too many permission prompts at startup
> 5-6 "Accept" clicks just to initialize. Rough first impression.

It's less for me for now, like 1-2, maybe I disabled something globally. But there's no way to disable specific questions (creating files, writing for certain folder etc) in cursor like you can do in claude. So it's asking everything, especially progress.json changes


> #### 12. "Skill" vs "Command" mode is weird for the user
> In Cursor, `/start` can be triggered as either a "skill" or a "command" and they behave differently:
> - **Command**: Wall of text, unclear how to proceed; gives full solution code directly (defeats the learning purpose)
> - **Skill**: Skips teaching, jumps straight to coding task

For me both works as expected. Probably you need to change cursorrules as I mentioned

> #### 13. Cursor command mode gives the full solution code directly
> Instead of teaching and letting the user code, Cursor in command mode just shows the complete solution code for the checkpoint. This defeats the entire learning purpose.

Can't reproduce, it seems for me both works the same

---

TODOs before merge (feel free to update this):
- [ ] test #399 and #400 properly
- [x] remove from READMEs current AI disabled block, add new AI block
- [ ] add AI icons for those challenges on SRE website
- [ ] present AI way on SRE website

CLAUDE.md
carletex, line 1: In SE-2 I think we are using AGENTS.md and linking it from here?

Also, would this be the file where we include the "non-teacher/guided mode" information? As discussed in https://github.com/scaffold-eth/se-2-challenges/issues/398

I could picture this file having the info about the challenge (or a linked file) and also presenting both modes (guided vs non-guided)

What do you guys think?

cc @technophile-04

technophile-04 (commented):
Minor nit on `.cursorignore.example`: the `node_modules/`, `.yarn/`, and `.git/` entries at the bottom are. Cursor already auto-ignores `.git/` and `node_modules/` by default, and also respects `.gitignore` (which covers `.yarn/`).

.cursorignore.example
```diff
@@ -0,0 +1,19 @@
+# SpeedRunEthereum Challenge - AI Configuration
+#
+# By default, AI features are ENABLED for the AI-guided learning path.
+# If you want to complete the challenge WITHOUT AI assistance (traditional mode),
+# uncomment the line below to block AI from accessing your files:
+#
+# *
```
technophile-04: This is a bit hidden. I didn't realise that here "*" is to actually ignore all the files. 

Maybe we just keep this file like this: 

```
# AI mode is ENABLED by default. To disable AI and do the challenge manually, uncomment the next line with "*" to disable:

# *
```

rin-st: Thanks, updated it with a small change: mentioned that it works only for cursor [25f4b78](https://github.com/scaffold-eth/se-2-challenges/pull/396/commits/25f4b78e1a37593ea3a094568d8f7cd633bc9b9c)

rin-st:
removed experience level and backmerged to #399 and #400

CLAUDE.md
rin-st: Removed CLAUDE.md file from the challenges, in the instances the one from [create-eth](https://github.com/scaffold-eth/create-eth/blob/main/templates/base/CLAUDE.md) is used, which refers to challenge's AGENTS.md. Added AI-guided part directly to challenge's AGENTS.md args file, example [[56051d6](https://github.com/scaffold-eth/se-2-challenges/pull/399/commits/56051d620e6be87a748f775f6a2e909b8ecf955b)](https://github.com/scaffold-eth/se-2-challenges/pull/399/commits/56051d620e6be87a748f775f6a2e909b8ecf955b)

cc @technophile-04

PR #393, opened 2026-01-23
initial iteration for general AI learning flow

technophile-04:
### To Test: 

1. Clone this example repo: 

```
git clone https://github.com/technophile-04/challenge-tokenization-with-ai.git
```

2. Open Cursor (in agent mode) or Calude code and just type: 

```
/start
```

### How it works: 

The `.ai/` dir: Contains the challenge content and AI behavior instructions. Shared by both Cursor and Claude Code and any future agents like opencode. 

`.ai/CHALLENGE.yaml`:  The master challenge definition file. Contains:
- `setup.template`: The TODO-marked contract template applied when `/start` is invoked
- `welcome_message`: Greeting shown at start
- `checkpoints[]`: Array of learning checkpoints, each with:
  - `context`: Teaching content shown BEFORE questions
  - `questions[]`: Questions with `concepts` array for evaluation and `hint` for help
  - `unlocks`: The code written to the contract when checkpoint completes
- `completion_message`: Shown when all checkpoints done

`.ai/instructions/start-content.md`: Detailed behavior instructions for the `/start` command. Tells the AI to:
- Apply the TODO template to YourCollectible.sol
- Initialize progress tracking
- Show welcome message
- Teach each checkpoint's context first, then ask questions
- Evaluate answers using concept matching
- Update the contract when checkpoints complete
- Use progressive hints when users are stuck

`.ai/instructions/continue-content.md`: Behavior instructions for the `/continue` command. Tells the AI to:
- Read existing progress
- Summarize completed checkpoints
- Resume from current position
- Follow same teaching approach as `/start`

`.challenge-ai/progress.json` - Runtime Progress (Gitignored)
Created at runtime when user runs `/start`. Tracks:
- `startedAt`: When the user began
- `currentCheckpoint`: Which checkpoint they're on
- `currentQuestion`: Which question within that checkpoint
- `checkpoints{}`: Status of each checkpoint (pending/in_progress/completed)

So basically it's like a save option in game. The AI auto-saves the user's progress in this file. 

---

The `.claude` dir

Here instead of creating the slash command we have skills, since now [claude depercated the slash commands](https://code.claude.com/docs/en/skills) and it comes udner skill only. 

`.claude/skills/start/SKILL.md`: Claude Code's `/start` skill:
- Points to `.ai/instructions/start-content.md` for detailed behavior
- Lists the startup steps (apply template, init progress, begin teaching)
- Body points to `.ai/instructions/start-content.md`

[...]

PR #397, opened 2026-02-11
test with crowdfund challenge

technophile-04:
### Description: 

I think the final goal would be #396, just creating this for inspiration. This shouldn't be merge, but was just tinkering with how #393 will work. Marking this is draft and we can close this later. 

--- 

So basically the flow is: 
1. AI gives the context about the challenge, what it is about and then what we will be doing. Tells to open the contract file which we will be editing (in this case crowd-funding) 
2. User says start, AI gives the context about checkpoint code and asks the user to implment the function in Solidity file
3. Once donce, users tells the AI to "check" and AI will run the corresponding test for that checkpoint if it fails AI will give more hints. 

To test: 

Clone the below repo and run `/start`

```
git clone https://github.com/technophile-04/crowd-funding-ai
```

<details>

<summary>
    Here is the test convo which I had: 
</summary>

```

 ▐▛███▜▌   Claude Code v2.1.39
▝▜█████▛▘  Opus 4.6 · Claude Max
  ▘▘ ▝▝    ~/Desktop/github/crowd-funding

❯ /start 

⏺ Read 2 files (ctrl+o to expand)

⏺ Now let me read the contract skeleton and initialize progress tracking in
  parallel.

⏺ Read 1 file (ctrl+o to expand)

⏺ progress-tracker(Initialize challenge progress)
  ⎿  Done (3 tool uses · 8.2k tokens · 17s)

⏺ Welcome to the Crowdfunding Challenge!

  You'll build a CrowdFund smart contract that lets users coordinate a group
  funding effort. If enough ETH is collected before a deadline, the funds go to
  a FundingRecipient contract. If not, everyone gets their money back. The users
   only have to trust the code, not each other.

  How this works:
  - I'll teach you a concept, then give you a coding task
  - You write the code in your editor, then say "check" and I'll run the tests
  - Say "hint" anytime you're stuck for progressive help
  - Use /skip if you want me to write the code for you (I'll still explain it!)

  Your progress is saved automatically. Use /start to resume anytime.

  Your contract is ready for you to start coding! I've opened it at
  packages/hardhat/contracts/CrowdFund.sol — you'll see empty function bodies
  waiting for your implementation. Let's fill them in together, one checkpoint
  at a time.

  ---
  Checkpoint 1: Contributing

  Let's learn three fundamental Solidity concepts you'll need for the
  contribute() function.

  Mappings: Your On-Chain Database

  A mapping is like a dictionary or hash map — it maps keys to values. In our
  case, we need to track how much each address has contributed:

  mapping(address => uint256) public balances;

  This creates a public lookup table: given any Ethereum address, you can check
  their balance. When you declare it public, Solidity automatically creates a
  getter function — anyone can call balances(0x123...) to check a specific
  address's contribution.

  Every address in a mapping starts at the default value (0 for uint256), so
  there's no need to initialize anything.

  Events: Broadcasting to the Outside World

  Smart contracts live on-chain, but frontends and indexing services live
  off-chain. Events are the bridge — they're log entries that get emitted during
   transactions and can be listened to by external applications.

  event Contribution(address contributor, uint256 amount);

  When you emit Contribution(msg.sender, msg.value), the frontend can detect
  this and update the UI in real-time. Without events, the frontend would have
  to constantly poll the contract to check for changes.

  Payable Functions and msg.sender/msg.value

  For a function to receive ETH, it must be marked payable. Inside any function,
   you have access to:
  - msg.sender — the address that called the function
  - msg.value — how much ETH (in wei) they sent with the call

  These are global variables available in every function — they come from the
  transaction itself.

  Putting It Together

  The contribute() function already exists in your contract but it's empty.
  You'll need to:
  1. Add a balances mapping to track contributions per address
  4. Add a Contribution event to notify the frontend
  5. Fill in contribute() to update the mapping and emit the event

  ---
  Does this make sense so far? Feel free to ask any questions about what you
  just read, or say "ready" when you want to continue!

✻ Crunched for 50s

❯ ready 

⏺ Great! Let's check your understanding with a quick question:

  If a user contributes twice — first 0.5 ETH then 0.3 ETH — what should their 
  balance be? And why do we use += instead of = when updating the mapping?

❯ the final balance would be 0.8 ETH and we use += so that we accumalte the 
   total                                                                    

⏺ Exactly right! If we used =, the second contribution of 0.3 ETH would
  overwrite the first one, and the user would lose track of their 0.5 ETH. Using
   += ensures we accumulate the total, so 0.5 + 0.3 = 0.8 ETH is correctly
  recorded.

  ---
  Your Task: Contributing

  Implement the contribute() function and add the supporting state and events.

  Open packages/hardhat/contracts/CrowdFund.sol and make these changes:

  1. Add a balances mapping under the State Variables section (below
  fundingRecipient):
  mapping(address => uint256) public balances;
  2. Add a Contribution event under the Events section:
  event Contribution(address, uint256);
  3. Fill in the contribute() function body to:
    - Add the sent amount to the sender's balance in the mapping
    - Emit the Contribution event with the sender's address and the amount

  When you're done, say "check" and I'll run the tests!
  Say "hint" if you need help, or /skip if you want me to write the code for
  you.

❯ check  

⏺ Read 1 file (ctrl+o to expand)

⏺ Bash(yarn test --grep 'Checkpoint1' 2>&1)
  ⎿  Error: Exit code 1
     Error HH606: The project cannot be compiled, see reasons below.

     The Solidity version pragma statement in these files doesn't match any of 
     the configured compilers in your config. Change the pragma or configure 
     additional compiler versions in your hardhat config.

       * contracts/CrowdFund.sol (0.8.20)
       * contracts/FundingRecipient.sol (0.8.20)

     To learn more, run the command again with --verbose

     … +4 lines (ctrl+o to see all)

⏺ Looks like there's a compiler configuration issue — the contract needs
  Solidity 0.8.20 but Hardhat isn't configured for it yet. Let me check the
  config.

⏺ Read 1 file (ctrl+o to expand)

⏺ The Hardhat config only has compiler 0.8.30 but the contracts use 0.8.20. Let
  me add the missing compiler version.

⏺ Update(packages/hardhat/hardhat.config.ts)
  ⎿  Added 9 lines
      35            },
      36          },
      37        },
      38 +      {                                                        
      39 +        version: "0.8.20",                                     
      40 +        settings: {                                            
      41 +          optimizer: {                                         
      42 +            enabled: true,                                     
      43 +            runs: 200,                                         
      44 +          },                                                   
      45 +        },                                                     
      46 +      },                                                       
      47      ],
      48    },
      49    defaultNetwork: "localhost",

⏺ Good, I fixed the compiler config. Let me run the tests again.

⏺ Bash(yarn test --grep 'Checkpoint1' 2>&1)
  ⎿  Warning: Function state mutability can be restricted to pure
       --> contracts/CrowdFund.sol:63:5:
        |
     … +57 lines (ctrl+o to expand)
  ⎿  (timeout 2m)

⏺ All 5 tests passed! Excellent work!

  Checkpoint Complete: Contributing!

  You nailed it — your implementation correctly:
  - Tracks balances per address using += to accumulate contributions
  - Emits a Contribution event so the frontend can react in real-time
  - Handles multiple contributors independently

  Nice use of the mapping + += pattern! This is the standard way to track
  per-user state in Solidity.

⏺ progress-tracker(Update checkpoint progress)
  ⎿  Done (2 tool uses · 8.4k tokens · 11s)

⏺ Want to see it in action? Run yarn deploy and check the frontend!

  When you're ready, let's move on to Checkpoint 2: Withdrawing Funds — where
  you'll learn about custom errors, the Checks-Effects-Interactions pattern, and
   how to safely send ETH back to users. Say "ready" to continue!

✻ Worked for 49s

❯ ready  

⏺ Checkpoint 2: Withdrawing Funds

  Great job on contribute()! Now let's handle the case where the funding fails
  and people need their money back.

  Custom Errors: Gas-Efficient Reverts

  In older Solidity, you'd use require(condition, "Error message") to validate
  inputs. Modern Solidity uses custom errors — they're more gas-efficient
  because error strings aren't stored on-chain:

  // Old way (costs more gas):
  require(isOpen, "Not open to withdraw");

  // New way (custom errors):
  error NotOpenToWithdraw();

  if (!isOpen) revert NotOpenToWithdraw();

  Custom errors can also carry data, which is useful for debugging:
  error WithdrawTransferFailed(address to, uint256 amount);

  Boolean State Flags

  We need a way to "unlock" withdrawals only when the funding campaign fails. A
  simple bool works perfectly:

  bool public openToWithdraw; // defaults to false

  Solidity initializes booleans to false, so withdrawals are locked by default.
  We'll flip this to true later (in Checkpoint 3) when the deadline passes
  without meeting the threshold.

  Sending ETH: The Low-Level Call

  To send ETH back to a user, we use a low-level call:

  (bool success, ) = recipient.call{value: amount}("");
  if (!success) revert TransferFailed();

  Why not transfer() or send()? Those forward only 2300 gas, which can cause
  failures with smart contract wallets. The call approach forwards all available
   gas and is the recommended pattern.

  The Checks-Effects-Interactions Pattern

  This is one of the most important security patterns in Solidity. When sending
  ETH, always:
  1. Checks — Validate conditions (is openToWithdraw true?)
  2. Effects — Update state BEFORE external calls (zero out the balance)
  3. Interactions — Make the external call (send ETH)

  Why? If you send ETH before zeroing the balance, a malicious contract could
  re-enter your withdraw() function and drain all funds. This is called a
  reentrancy attack — it's how the famous DAO hack happened.

  // SAFE: Checks-Effects-Interactions
  uint256 amount = balances[msg.sender];
  balances[msg.sender] = 0;           // Effect: zero out FIRST
  (bool success, ) = msg.sender.call{value: amount}("");  // Interaction: send
  AFTER

  ---
  Does this make sense so far? Feel free to ask questions, or say "ready" when
  you want to continue!
```

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,21 @@ yarn start
 
 ---
 
+## 🤖 AI-Guided Learning Mode (Optional)
+
+Want an interactive tutor that teaches you the concepts while you code? This challenge supports **AI-guided learning mode**!
+
+1. Open this project in **Claude Code** or **Cursor**
+2. Run the `/start` command
+3. The AI tutor will teach you each concept, then give you a coding task
+4. You write the code, say **"check"**, and the AI runs the tests
+5. Say **"hint"** for help, or **`/skip`** if you want the AI to show you the solution
+6. Your progress is saved — use `/start` to resume anytime
+
+The AI won't just give you the answers — it teaches first, then has you implement the code yourself. Tests validate your work, and the AI helps you debug if something doesn't pass.
+
+---
+
 ## 🧑‍🚀 Your Mission
 
 🏦 Build a `CrowdFund.sol` contract that collects **ETH** from numerous addresses using a payable `contribute()` function and keeps track of `balances`. After some `deadline` if it has at least some `threshold` of ETH, it sends it to a `FundingRecipient` contract (This is a stand-in for any potential use case a group of people would want to fund together). It then triggers the `complete()` action, sending the full balance. If not enough **ETH** is collected, allow users to `withdraw()`.
--- a/extension/.ai/CHALLENGE.yaml
+++ b/extension/.ai/CHALLENGE.yaml
@@ -1,0 +1,525 @@
+name: "Crowdfunding App"
+version: "1.0"
+description: "Build a decentralized crowdfunding contract where users coordinate group funding. Learn about mappings, events, payable functions, custom errors, state machines, and the Checks-Effects-Interactions pattern."
+difficulty: "intermediate"
+estimated_time: "60-90 minutes"
+
+# No setup.template — the contract already has a skeleton with empty function bodies.
+# The user fills them in progressively as they complete each checkpoint.
+
+welcome_message: |
+  Welcome to the Crowdfunding Challenge!
+
+  You'll build a `CrowdFund` smart contract that lets users coordinate a group funding effort.
+  If enough ETH is collected before a deadline, the funds go to a `FundingRecipient` contract.
+  If not, everyone gets their money back. The users only have to trust the code, not each other.
+
+  **How this works:**
+  - I'll teach you a concept, then give you a coding task
+  - You write the code in your editor, then say **"check"** and I'll run the tests
+  - Say **"hint"** anytime you're stuck for progressive help
+  - Use **`/skip`** if you want me to write the code for you (I'll still explain it!)
+
+  Your progress is saved automatically. Use `/start` to resume anytime.
+
+  Let's build something awesome!
+
+checkpoints:
+  # ============================================================
+  # CHECKPOINT 1: Contributing
+  # ============================================================
+  - id: "contributing"
+    title: "Contributing"
+    description: "Track contributions with mappings, emit events, and implement the contribute() function"
+
+    context: |
+      ## Mappings, Events, and Payable Functions
+
+      Let's learn three fundamental Solidity concepts you'll need for the `contribute()` function.
+
+      ### Mappings: Your On-Chain Database
+
+      A **mapping** is like a dictionary or hash map — it maps keys to values. In our case, we need to track how much each address has contributed:
+
+      ```solidity
+      mapping(address => uint256) public balances;
+      ```
+
+      This creates a public lookup table: given any Ethereum address, you can check their balance.
+      When you declare it `public`, Solidity automatically creates a getter function — anyone can call `balances(0x123...)` to check a specific address's contribution.
+
+      Every address in a mapping starts at the default value (`0` for uint256), so there's no need to initialize anything.
+
+      ### Events: Broadcasting to the Outside World
+
+      Smart contracts live on-chain, but frontends and indexing services live off-chain. **Events** are the bridge — they're log entries that get emitted during transactions and can be listened to by external applications.
+
+      ```solidity
+      event Contribution(address contributor, uint256 amount);
+      ```
+
+      When you `emit Contribution(msg.sender, msg.value)`, the frontend can detect this and update the UI in real-time. Without events, the frontend would have to constantly poll the contract to check for changes.
+
+      ### Payable Functions and msg.sender/msg.value
+
+      For a function to receive ETH, it must be marked `payable`. Inside any function, you have access to:
+      - **`msg.sender`** — the address that called the function
+      - **`msg.value`** — how much ETH (in wei) they sent with the call
+
+      These are global variables available in every function — they come from the transaction itself.
+
+      ### Putting It Together
+
+      The `contribute()` function already exists in your contract but it's empty. You'll need to:
+      1. Add a `balances` mapping to track contributions per address
+      2. Add a `Contribution` event to notify the frontend
+      3. Fill in `contribute()` to update the mapping and emit the event
+
+    questions:
+      - text: "If a user contributes twice — first 0.5 ETH then 0.3 ETH — what should their balance be? And why do we use `+=` instead of `=` when updating the mapping?"
+        concepts: ["accumulate", "add", "0.8", "plus", "previous", "total", "sum", "overwrite"]
+        hint: "Think about what happens if you use `=` instead of `+=`. The second contribution would overwrite the first one!"
+
+    task:
+      description: |
+        Implement the `contribute()` function and add the supporting state and events.
+
+        Open `packages/hardhat/contracts/CrowdFund.sol` and make these changes:
+
+        1. **Add a `balances` mapping** under the State Variables section (below `fundingRecipient`):
+           ```solidity
+           mapping(address => uint256) public balances;
+           ```
+
+        2. **Add a `Contribution` event** under the Events section:
+           ```solidity
+           event Contribution(address, uint256);
+           ```
+
+        3. **Fill in the `contribute()` function** body to:
+           - Add the sent amount to the sender's balance in the mapping
+           - Emit the `Contribution` event with the sender's address and the amount
+
+        When you're done, say **"check"** and I'll run the tests!
+      file: "packages/hardhat/contracts/CrowdFund.sol"
+      test: "yarn test --grep 'Checkpoint1'"
+      hints:
+        - "Use `msg.sender` for the caller's address and `msg.value` for the ETH amount they sent. Remember to use `+=` to accumulate contributions!"
+        - |
+          Your `contribute()` body should have two lines:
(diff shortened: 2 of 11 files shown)


Pabl0cks (commented):
I've ran through the whole experience in Claude mainly, because Cursor was not working great (I think because the `.cursorignore` rules). Here is the initial report:

## Personal view

- The Claude teaching flow, Q&A, progressive hints, and test validation all worked smoothly
- The code reviews and the "teacher" explanations on what your errors where and how to solve them were amazing. REALLY nice experience to have a personal teacher seating with you.
- The progress was stored really nice, cursor could see nicely in which step of the challenge I was (I started a new conv after completing steps on Claude and it detected nicely where I was)
- The "reading" experience in Claude terminal is a bit harsh, much nicer to read in Cursor with nice styles
- The experience of going to Claude/Cursor still may have too much friction for "builders", having a web UI is on the roadmap?
- Antigravity does not support this yet, right?

## AI Report based on my internal notes

The AI redacting might express too strong opinions in some suggestion!

### Bugs / Broken Stuff

#### 1. `yarn install` not run before first test
First "check" fails — `node_modules` doesn't exist. Should run `yarn install` during setup or warn the user beforehand.

#### 2. Hardhat config missing Solidity 0.8.20 compiler
Contract uses `0.8.20` but config only has `0.8.30`. First compile fails with `HH606`. Should be pre-configured in the repo.

#### 3. `.cursorignore` blocks contract file access in Cursor
Cursor can't read `CrowdFund.sol` — filtered by `.cursorignore`. Breaks the whole Cursor experience.

#### 4. `yarn generate` / `yarn deploy` need interactive input
Both prompt for passwords — crashes when the AI tries to run them. Should tell users to run these in their own terminal.

### UX Issues (Claude Code)

#### 5. User doesn't know WHERE to write code
The very first reaction was "Not clear to me where do I need to write what." For someone using Claude Code (terminal only), being told "open packages/hardhat/contracts/CrowdFund.sol" might not be enough. Maybe it should hint something?

#### 6. "hint" not visible after long teaching blocks
Welcome message mentions "hint" but it scrolls off-screen after the big teaching wall. The closing prompt only says `say "ready"` — should also mention "hint", especially on that first overwhelming checkpoint.

#### 7. Checkpoint 4 task description gives away the answer
Task says "just call `contribute()` from inside `receive()`" — that's literally the solution. User called it out. Make the description vaguer and let them figure it out.

#### 8. progress.json prompt confuses beginners
Users see a big red/green diff and "Do you want to overwrite progress.json?" with no context. Consider auto-allowing progress file writes.

#### 9. Completion message says `yarn deploy` without a network
Deploys to localhost. Should say `yarn deploy --network sepolia`. (The AI caught this in conversation when asked — it's a CHALLENGE.yaml template issue.)

#### 10. Suggested faucets require mainnet balance
Sepoliafaucet.com and Alchemy faucets need mainnet ETH. Use Google's instead: https://cloud.google.com/application/web3/faucet/ethereum/sepolia

---

### UX Issues (Cursor). Probably most are caused because the `cursorignore`?

#### 11. Too many permission prompts at startup
5-6 "Accept" clicks just to initialize. Rough first impression.

#### 12. "Skill" vs "Command" mode is weird for the user
In Cursor, `/start` can be triggered as either a "skill" or a "command" and they behave differently:
- **Command**: Wall of text, unclear how to proceed; gives full solution code directly (defeats the learning purpose)
- **Skill**: Skips teaching, jumps straight to coding task

#### 13. Cursor command mode gives the full solution code directly
Instead of teaching and letting the user code, Cursor in command mode just shows the complete solution code for the checkpoint. This defeats the entire learning purpose.

---

<details><summary>Cursor screenshots </summary>

First run, doesn't look like got the context right because the cursorignore:
image

Running the command:
image

Running the skill:
image

</details>

rin-st:
Yes, to make cursor work you need to remove `.cursorignore` file and also change vscode settings file like [this](https://github.com/scaffold-eth/se-2-challenges/pull/396/changes#diff-fe366aa32fc7dfc84704d3e32c43d2b7cc52356cc69bf409d306abf312c72d59) so cursor won't suggest you solutions when you type

For me `/start` command in cursor works as it should. But yes, too much requests for permissions for now.

rin-st:
Forgot to say, I'll fix what's possible from https://github.com/scaffold-eth/se-2-challenges/pull/397#pullrequestreview-3784538560 report in #396 and related PRs. Thanks, Pablo!

PR #400, opened 2026-02-13
Challenge crowdfunding - AI guided

rin-st:
```
npx create-eth@2.0.10 -e https://github.com/scaffold-eth/se-2-challenges/tree/challenge-ai-crowdfunding
```

Flow is almost the same as in #397 but challenge recreated from `learn-with-ai-combined` - `base-challenge-template` with AI changes. The main difference from `learn-with-ai-combined` is `CHALLENGE.yaml` file which describes behavior.

Also, updated the challenge to create-eth@2.0.10 so it will be easier to make next changes when needed


technophile-04 (approved):
Working great, Thanks @rin-st !! Looks like something we can already ship! 

Some views on difficulty: 
I tried the intermediate and advanced mode: 
My expectation of Advanced mode was that it would teach me more about the gas optimization, security tricks, or tell me more about edge cases, etc. But it was just more strict and providing to the point context about the checkpoint not too much talking, which felt like advanced actually and make sense to have it guardrailed by challenge content but maybe in advanced mode maybe we can decrease the temprature and ask it to be creative. But this is for fututre iteration! 

Maybe the next steps would: 
Figuring out how are gonna pitch it to the outter world? Maybe an Line in the README? or on the website that if you are following the challenges you can use AI mode?  

Another future idea: 
We could also have a `/reveiw` command. Which will help you review / revise the whole challenge nicely. Since `progresss.json` tracks the history of the skipped checpoint we can ask AI to focus on skipped parts more as well

PR #413, opened 2026-02-24
add predictionNotReported modifier hint to Checkpoint 5 solution code

phipsae:
Hi, just two quick fixs:

The solution code for Checkpoint 5 was missing guidance to add the predictionNotReported modifier to addLiquidity and removeLiquidity. Because of that, students missed this step and failed the corresponding tests.

There is also a formatting issue in one of the headers, which is now fixed.


rin-st (approved):
Lgtm, thank you!

PR #414, opened 2026-02-24
Challenge zk voting

phipsae:
Hi there, 

I completely forgot to add a checkpoint for contract verification. Here it is :)


rin-st (approved):
Lgtm, thank you!

PR #416, opened 2026-02-26
Challenge dex - AI guided

rin-st:
```
npx create-eth@2.0.10 -e https://github.com/scaffold-eth/se-2-challenges/tree/challenge-dex-ai
```


Pabl0cks (approved):
Awesome job @rin-st !! Worked nicely.

A nitpick that may apply to more challenges, can totally ignore or improve in a future PR or when we implement the web AI tutor:

When we are initializing the challenge, we receive all the info about the "3 commands", and the AI guided mechanics, how to redeploy, and then we jump straight to Challenge 1.

There is a lot of info there, you have to scroll up in Claude code, maybe we could ask to "continue" or something after the first chunk of text (after `Re-run yarn deploy --reset ...")

Image

technophile-04:
Great job @rin-st! This seems to perf! 

> When we are initializing the challenge, we receive all the info about the "3 commands", and the AI guided mechanics, how to redeploy, and then we jump straight to Challenge 1.
> 
> There is a lot of info there, you have to scroll up in Claude code, maybe we could ask to "continue" or something after the first chunk of text (after `Re-run yarn deploy --reset ...")

This is really nice add! I think we should add it, yup maybe just after giving intro about the challenges and the 3 terminal commands. We tell AI to pause ask the user to say "ready" so that it starts with checkpoint.

rin-st:
Good points! Now it asks for open terminals and texting "ready" after "How it works" block. I think after "How it works" is a bit better than after yarn chain/deploy/start since
1. after "ready" user starts right from first checkpoint
2. it makes "how it works" block more noticeable

backmerged the change to all the challenges, but can change if needed

image

PR #420, opened 2026-02-26
Challenge prediction markets - AI guided

rin-st:
[AI-guided learning system](https://github.com/scaffold-eth/se-2-challenges/pull/396#issue-3921710005) for oracles challenge

To install

```
npx create-eth@2.0.10 -e https://github.com/scaffold-eth/se-2-challenges/tree/challenge-prediction-markets-ai
```

and then from claude / cursor

```
/start
```


extension/.ai/CHALLENGE.yaml
```diff
@@ -0,0 +1,886 @@
...
+      uint256 public s_lpTradingRevenue;   // fees from trading
+      ```
+
+      ### Validations
+      - `msg.value` must be > 0 (need initial liquidity)
+      - `_initialYesProbability` must be > 0 and < 100
+      - `_percentageToLock` must be > 0 and < 100
+
+    questions:
+      - text: "Why must the initial probability be strictly between 0% and 100%? What would happen at the extremes?"
```
phipsae: The content we are giving doesnt allow the user to answer this question.

extension/.ai/CHALLENGE.yaml
```diff
@@ -0,0 +1,886 @@
...
+      - Starting probability = 12 / (12 + 8) = 60%
+
+      ### Why Lock Tokens?
+
+      Without locking, a single buy pushes probability to 100%. Locking provides price stability from the start.
+
+      After implementing, uncomment the code below `/// Checkpoint 3 ///` in `getPrediction()`.
+
+    questions:
+      - text: "Why do we lock tokens to the owner instead of just setting a probability variable? What advantage does this approach have?"
```
phipsae: The context above teaches "price stability" as the reason for locking, but the question expects an answer about pricing engine consistency. 

Image

extension/.ai/CHALLENGE.yaml
```diff
@@ -0,0 +1,886 @@
...
+      2. Collect accumulated `s_lpTradingRevenue`
+      3. Burn the winning tokens
+      4. Send total ETH to the LP
+
+      You need a `predictionReported` modifier (opposite of `predictionNotReported`).
+
+      After implementing report, uncomment code below `/// Checkpoint 5 ///` in `getPrediction()`.
+
+    questions:
+      - text: "Why do we need the predictionNotReported modifier on addLiquidity and removeLiquidity? What could go wrong without it?"
```
phipsae: There's no manipulation or cheating scenario here — the LP can't gain an unfair advantage by adding/removing liquidity after reporting. The real answer is simpler: we just disable it because it doesn't make sense to change liquidity after the outcome is decided. The expected concepts (manipulat, unfair, cheat, game) are misleading.

extension/.ai/CHALLENGE.yaml
```diff
@@ -0,0 +1,886 @@
...
+      4. **Average** the before/after probabilities
+      5. **Final price**: `(initialTokenValue * probabilityAvg * tradingAmount) / (PRECISION * PRECISION)`
+
+      For buying: reserve decreases → more tokens "sold" → probability goes up
+      For selling: reserve increases → fewer tokens "sold" → probability goes down
+
+      `getBuyPriceInEth` and `getSellPriceInEth` are thin wrappers calling `_calculatePriceInEth` with a `_isSelling` flag.
+
+    questions:
+      - text: "Why do we average the probability before and after the trade instead of just using the current probability?"
```
phipsae: We dont give any context to answer the question to the user

extension/.ai/CHALLENGE.yaml
```diff
@@ -0,0 +1,886 @@
...
+      ### sellTokensForEth
+
+      - Check user has enough tokens AND has approved the contract
+      - Calculate ETH with `getSellPriceInEth`
+      - Subtract from `s_lpTradingRevenue`
+      - Send ETH to seller, pull tokens back via `transferFrom`
+      - Emit `TokensSold`
+
+    questions:
+      - text: "Why can't the market owner buy or sell tokens? What unfair advantage would they have?"
```
phipsae: There is non really, we just clearly separate the roles in the challenge. If the owner wants to trade he can even do that with another account.

extension/.ai/CHALLENGE.yaml
```diff
@@ -0,0 +1,886 @@
...
+      3. Calculate ETH: `(_amount * i_initialTokenValue) / PRECISION`
+      4. Subtract from `s_ethCollateral`
+      5. Burn the winning tokens
+      6. Send ETH to user
+      7. Emit `WinningTokensRedeemed`
+
+      The `notOwner` modifier prevents the LP from redeeming (they use `resolveMarketAndWithdraw` instead).
+
+    questions:
+      - text: "Why do we burn the winning tokens during redemption rather than just transferring them back to the contract?"
```
phipsae: Same issue, we dont provide the context to the question

phipsae (changes requested):
Thanks for putting this together. A few thoughts on the format and direction:

I found the challenge somewhat difficult to consume in its current form. The terminal-based tutorial approach doesn't land as well for me compared to the existing SRE webpage format, which presents the material more clearly. Even using /skip to bypass writing code, I'd still prefer working through the challenge directly on the site. For just conveying the concepts (someone using just /skip) we might also just consider an article.

I do like the inclusion of questions and also the possibility to ask questions directly inside the challenge. Maybe integrating a chat window directly into the SRE challenge page could be an alternative. 

On the frontend side, it feels largely absent. The agent suggests jumping into it at some checkpoints, but those prompts are easy to overlook, and it's unclear whether users would actually follow through on them.

Image
￼
For Checkpoint 2 the agent forgot about it to mention. 

Image

￼
Even if frontends become less relevant going forward, having learners tinker with the frontend in SRE is a good way to help them understand the protocols better.

These are just my impressions/feelings, let me know what you think.

rin-st:
Thank you for review!

> I found the challenge somewhat difficult to consume in its current form. The terminal-based tutorial approach doesn't land as well for me compared to the existing SRE webpage format, which presents the material more clearly. Even using /skip to bypass writing code, I'd still prefer working through the challenge directly on the site. For just conveying the concepts (someone using just /skip) we might also just consider an article.

Yep, it's not ideal for now. And wall of text is one of the main disadvantages. We'll keep iterating and hopefully will fix it. And, AI-guided mode is just an option, not a must. Anyone could use the mode they want to use

> I do like the inclusion of questions and also the possibility to ask questions directly inside the challenge. Maybe integrating a chat window directly into the SRE challenge page could be an alternative.

Yep, asking questions was one of the reasons to add this mode. And we're thinking about adding chat window too

> On the frontend side, it feels largely absent. The agent suggests jumping into it at some checkpoints, but those prompts are easy to overlook

I updated frontend related parts and now all (or at least most) of the frontend interactions should be mentioned by AI. But I can't use images in CLI, so yeah, they're absent

> and it's unclear whether users would actually follow through on them.

Well, it's difficult to track it on our side. It's not clear in regular mode too

---

I updated all the questions you mentioned in your comments. Now all the questions should be based on the given context [1ef009d](https://github.com/scaffold-eth/se-2-challenges/pull/420/commits/1ef009d59e5d9c0b2820aff275b9df1716b6ea26)

rin-st:
Merging this for now, please lmk if we need to change something

PR #421, opened 2026-02-26
Challenge zk voting - AI guided

rin-st:
[AI-guided learning system](https://github.com/scaffold-eth/se-2-challenges/pull/396#issue-3921710005) for oracles challenge

To install

```
npx create-eth@2.0.10 -e https://github.com/scaffold-eth/se-2-challenges/tree/challenge-zk-voting-ai
```

and then from claude / cursor

```
/start
```


phipsae:
Thanks for the challenge, it works well. The same feedback from the prediction market challenge applies here too :)

I’m going to open another PR, because the BB installation no longer works as described.

PR #427, opened 2026-03-10
Challenge zk voting

phipsae:
Hi,        

The `bbup` installer script was unreliable, so I replaced it with direct binary download instructions for each platform (Apple Silicon, Intel Mac, Linux) in both README.md and extension/README.md.args.mjs.
                                                                                                                        
Would be great if you could test whether the installation works for you.

At some point I should also update the challenge to the latest versions of Noir and bb.

Thanks and best
Philip

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,16 @@ nargo --version
 
 Install with:
 
-```sh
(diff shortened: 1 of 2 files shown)


rin-st:
Apple silicon worked for me
Added small change, js -> sh in readme.md.args.mjs
Merging, thanks!

PR #429, opened 2026-03-20
Challenge tokenization foundry

rin-st:
- universal start-content and skip-content for all the challenges #430 

This challenge
- universal page.tsx
- universal challenge.yaml 
- Readme.md.args.mjs and Agents.md.args.mjs depends on SF. So in the instance resulting README.md and AGENTS.md have only chosen SF part
- Main README.md, which is used on SRE website have tabs for hardhat and foundry

foundry part
- contracts
- deploy scripts
- test file (which will be different from the grader hh version)

⚠️ Merge after https://github.com/BuidlGuidl/SpeedRunEthereum-v2/pull/367

PR #431, opened 2026-03-23
Challenge crowdfunding foundry

rin-st:
Changes are very similar to https://github.com/scaffold-eth/se-2-challenges/pull/429


technophile-04:
Is this ready?

rin-st:
> Is this ready?

It wasn't but now it's ready

extension/.ai/CHALLENGE.yaml
```diff
@@ -142,10 +144,12 @@ checkpoints:
...
 
-        > Need to troubleshoot? `hardhat/console.sol` is already imported so you can use `console.log()` in your Solidity code. Output appears in your `yarn chain` terminal.
+        > Need to troubleshoot? You can use `console.log()` in your Solidity code. Output appears in your `yarn chain` terminal.
 
         When you're done, say **"check"** and I'll run the tests!
-      file: "packages/hardhat/contracts/CrowdFund.sol"
+      file:
+        hardhat: "packages/hardhat/contracts/CrowdFund.sol"
+        foundry: "packages/foundry/contracts/CrowdFund.sol"
       test: "yarn test --grep 'Checkpoint1'"
```
Pabl0cks: I think in the tests we should also specify hardhat/foundry, if not it keeps trying the `yarn test --grep 'Checkpoint1'` and failing, instead of `yarn test --match-test 'Checkpoint1'`

rin-st: Good point Pablo, thanks!

Fixed it, [9aad814](https://github.com/scaffold-eth/se-2-challenges/pull/430/commits/9aad8143e1b9b31709bdbe741081d981b83d780f) at #430 and backmerged here, and [ba70c62](https://github.com/scaffold-eth/se-2-challenges/pull/431/commits/ba70c624eb6be7214c599e6e892e012cc78274c2) in this PR

The main issue: foundry compiles all the tests even when running `--match-test Checkpoint1`, so if some functions is not yet implemented, tests file when compiling. Fix: `ICrowdFund` interface with all the functions and typecast `ICrowdFund(...)` in test file and now everything works as expected. [--match-test docs](https://www.getfoundry.sh/config/reference/testing#match-test)

Other solutions:
- make empty functions/states in every initial challenge contract. It can break the flow of some challenges (including crowdfunding) and requires rewriting contracts/readmes

Tried also to break test file to multiple files by checkpoints e.g. `CrowdFundCheckpoint1` and then use [--match-path](https://www.getfoundry.sh/config/reference/testing#match-path) flag which could be great solution, but unfortunately when running it foundry tries to compile all the files instead of just one

cc @technophile-04

technophile-04: Doing some research I feel the interface pattern which you added now is really simpler and nice. With the `--match-path` we can actually make it work by using the `--skip` so it would be like `forge test --match-path test/CrowdFundCheckpoint1.t.sol --skip Checkpoint2 --skip Checkpoint3 --skip Checkpoint4` 

But then we need to update other the `CHALLENGE.yaml` + also the testing instruciton in README. I feel current approach is best 🙌, Thanks!

extension/.ai/CHALLENGE.yaml
```diff
@@ -297,7 +301,9 @@ checkpoints:
         > **Important:** Once you're done testing, switch `openToWithdraw` back to its default (`bool public openToWithdraw;`) before running the tests!
 
         Say **"check"** when you're ready!
-      file: "packages/hardhat/contracts/CrowdFund.sol"
+      file:
+        hardhat: "packages/hardhat/contracts/CrowdFund.sol"
+        foundry: "packages/foundry/contracts/CrowdFund.sol"
       test: "yarn test --grep 'Checkpoint2'"
```
technophile-04: So now here we also need to mention different foundry command: 

```
test:
  hardhat: "yarn test --grep 'Checkpoint1'"
  foundry: "yarn test --match-test Checkpoint1"
```

extension/.ai/CHALLENGE.yaml
rin-st: I tried this `forge test --skip Checkpoint2 --skip Checkpoint3 --skip Checkpoint4` when testing and it worked, but I didn't add it to my comment since I didn't find `--skip` flag in the docs and yeah it's ugly to write tests that way and update other related files, so just skipped `--skip` 🙂

extension/.ai/CHALLENGE.yaml
rin-st: weird, I was sure I changed it. Probably changed again occasionally when force pushed. Thanks!

[e65acf3](https://github.com/scaffold-eth/se-2-challenges/pull/431/commits/e65acf3f812b4d7a22e26d749e065e0559a70dfa)

carletex (approved):

damianmarti:
@rin-st Sorry, I was testing this before it was merged. It worked great, except for a small issue with the AI-guided. The AI didn't mention anything about the side quest to avoid the funds from getting trapped in the contract. I think this is an important issue, and the user should be warned about it by the AI in some way. The message from the AI was that the challenge was completed and information about the contract like it was finished and secure.

rin-st:
As I remember for me AI mentioned that. Created an issue, will check again later. Thanks!
https://github.com/scaffold-eth/se-2-challenges/issues/450

PR #433, opened 2026-03-27
Challenge token vendor foundry

rin-st:
## Summary
- Add Foundry contracts, deploy scripts, and tests for the Token Vendor challenge
- Update CHALLENGE.yaml with dictionary file paths for both frameworks
- Update AGENTS.md.args.mjs and README.md.args.mjs with `solidityFramework` parameter
- Add tabs to README.md for framework-specific instructions
- Add AI-Guided Learning Mode section to README.md.args.mjs

## Test plan
- [ ] Verify foundry contracts compile
- [ ] Run `yarn test` with foundry framework
- [ ] Verify README.md renders tabs correctly
- [ ] Verify README.md.args.mjs renders correctly for both hardhat and foundry
- [ ] Verify AGENTS.md.args.mjs renders correctly for both frameworks
- [ ] Verify CHALLENGE.yaml file paths resolve for both frameworks

🤖 Generated with [Claude Code](https://claude.com/claude-code)


extension/packages/foundry/test/Vendor.t.sol
```diff
@@ -0,0 +1,207 @@
...
+
+        assertEq(yourToken.balanceOf(user), startingBalance + expectedTokens);
+    }
+
+    function test_Checkpoint2_BuyTokensEmitsBuyTokensEvent() public {
+        uint256 ethToSpend = 0.1 ether;
+        uint256 expectedTokens = ethToSpend * TOKENS_PER_ETH;
+
+        vm.expectEmit(true, false, false, true);
+        emit IVendor.BuyTokens(user, ethToSpend, expectedTokens);
```
technophile-04: This was failing for me: 

```
yarn test --match-test "Checkpoint2"
Compiler run failed:
Error (9582): Member "BuyTokens" not found or not visible after argument-dependent lookup in type(contract IVendor).
  --> test/Vendor.t.sol:84:14:
   |
84 |         emit IVendor.BuyTokens(user, ethToSpend, expectedTokens);
   |              ^^^^^^^^^^^^^^^^^

Error: Compilation failed
```

talking with AI it says that, this syntax is not allowed in 0.8.20 and we have to move the events from Interface to this test file itself. If our contract was above 0.8.21 then this would've worked but since our contract is 0.8.20 the whole test suite uses that version. 

Moving those events declaration to this file it passed.

technophile-04 (commented):
Thanks @rin-st! This is working nicely! just mentioend one small comment. 

--- 

Also maybe not related to this challenge, but when switching the tabs (the header links). For some Scaffold-UI components instead of taking SRE colors it fallbacks to dark blue colors. 

Image

extension/packages/foundry/test/Vendor.t.sol
rin-st: [ea5ea0b](https://github.com/scaffold-eth/se-2-challenges/pull/433/commits/ea5ea0bdd71766c38056d7fbd649d329e8db8919) updated, thanks!

rin-st:
> Also maybe not related to this challenge, but when switching the tabs (the header links). For some Scaffold-UI components instead of taking SRE colors it fallbacks to dark blue colors

Unfortunately it's working that way for all the challenges. Not sure when it broke. Created the issue:

https://github.com/scaffold-eth/se-2-challenges/issues/447

technophile-04 (approved):
Thanks @rin-st! Tested it out with applying the readme solution and everything works as expected!

--- 

> Unfortunately it's working that way for all the challenges. Not sure when it broke. Created the issue:
>
> https://github.com/scaffold-eth/se-2-challenges/issues/447

Ohhh I seee, maybe not the priority but we shall solve it as sooon as we get time 🙌 Will try to also debug what's wrong. But I remmber we having this similar issue and remmber we fixing it for challenges but maybe I am hallucinating 

---- 

Also do we need to update our notions solution link? I copied the solution from there but the test failed. But I think that solution is old? and we updated the challenge test to match some exact custom error names? That old solution will also fail with latest hardhat I feel. Talking with AI it said: 


Why your old solution failed both

The old solution predates the tightening of the grader. Both the Hardhat and Foundry
tests were rewritten to check:
1. Exact custom-error names (InvalidEthAmount, InsufficientVendorTokenBalance,
InvalidTokenAmount, InsufficientVendorEthBalance, EthTransferFailed).
2. Exact event shapes — SellTokens(indexed seller, amountOfTokens, amountOfETH). Your
 old event had amountOfETH and amountOfTokens swapped, which breaks .withArgs
regardless of framework.
3. Pre-condition checks firing before external calls (the "insufficient vendor ETH"
test proves the revert happens before transferFrom touches state).
4. A specific EthTransferFailed(address, uint256) from withdraw — which the Hardhat
suite verifies via the clever "transfer ownership to the vendor itself so call fails"
 trick, and which your old Vendor__WithdrawError() can't satisfy.

rin-st:
> Also do we need to update our notions solution link?

Updated some solutions. Sources

Oracles - https://github.com/scaffold-eth/se-2-challenges/pull/377#issue-3576085789
OCL - https://github.com/scaffold-eth/se-2-challenges/pull/351#issue-3319745232
Token vendor - AI generated from the challenge Readme

---

Now we can always generate solution from the challenge instructions. But let's keep solutions just in case

PR #443, opened 2026-04-02
feat: add CONCEPTS.yaml for crowdfunding challenge

technophile-04:
## Summary
- Adds `extension/.ai/CONCEPTS.yaml` for the crowdfunding challenge web chatbot (Phase 1 concept delivery)
- 4 checkpoints: trustless coordination → tracking contributions → safe withdrawals (reentrancy/CEI) → state machines & deadlines
- Each checkpoint has 2 open-ended questions with concept keywords and hints, all audited to only test content from their own checkpoint's context

## Test plan
- [ ] Verify CONCEPTS.yaml is picked up by `fetchGithubConceptsYaml()` in the SRE-v2 web app
- [ ] Walk through all 4 checkpoints in the chatbot — concepts should flow naturally and questions should be answerable from the checkpoint context alone
- [ ] Confirm Phase 1 → Phase 2 transition works (bot offers local setup after last checkpoint)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/extension/.ai/CONCEPTS.yaml
+++ b/extension/.ai/CONCEPTS.yaml
@@ -8,7 +8,7 @@ welcome_message: |
   We'll walk through the key concepts first: how contributions are tracked on-chain, why sending ETH back safely is trickier than you'd expect, and how the contract decides whether the campaign succeeded or failed. Once you've got the mental model, I'll help you set it up locally and start coding.
 
 completion_message: |
-  You now understand the core mechanics behind on-chain crowdfunding — how mappings track who gave what, why the Checks-Effects-Interactions pattern exists, how state machines model contract logic, how contracts handle unexpected ETH, and why thinking about edge cases is just as important as writing the happy path.
+  You now understand the core mechanics behind on-chain crowdfunding — how on-chain tracking works, why the order of operations matters when moving ETH, how state machines model contract logic, and why thinking about edge cases is just as important as the happy path.
 
   Let me know when you're ready to set up the challenge locally and I'll walk you through it step by step.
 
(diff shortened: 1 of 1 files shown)


extension/.ai/CONCEPTS.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+      Once you have the core flow working, the next question is: what could go wrong? This is where smart contract thinking differs from regular programming. What happens if someone sends ETH *after* the campaign already succeeded and the funds were forwarded? That ETH is now stuck — the campaign is over, there's no mechanism to move it out. Good contract design uses guards (like modifiers) to prevent functions from running when the contract is in a state where they shouldn't. The goal is to make invalid states impossible, not just unlikely.
+    questions:
+      - question: "After the deadline passes and not enough ETH was collected, why can't contributors just withdraw directly? Why does someone need to trigger the state change first?"
+        concepts: ["state transition", "trigger", "someone", "can't auto-execute", "no automatic", "transaction", "enable"]
+        hint: "Think about how smart contracts execute code — can they do anything on their own, without someone sending a transaction?"
+      - question: "What happens if someone sends ETH to the contract after the campaign has already succeeded and funds were forwarded? Why is this a problem?"
```
rin-st: nit: it could be difficult to answer this question  especially when AI asks you about the details after simple answer like "It will be stuck" because
- theory is somewhere before first question ("Think about how smart contracts execute code...") and discussion about answers
- it's not obvous why it's impossible to move it out user is not reading the contract

Related to all the second questions in the checkpoints though. But I don't know how to make it better, maybe give related theory right before the question? I think let's keep it as is for now and see how it goes for users

rin-st (approved):
Great concepts! It was a bit more difficult to answer than tokenization, added one related comment

But overall LGTM, thank you!

extension/.ai/CONCEPTS.yaml
technophile-04: [`e221fa4` (this PR)](https://github.com/scaffold-eth/se-2-challenges/pull/443/commits/e221fa4e697302adc8c755b76f8b2214f8f19b98) split the what + why style questions in checkpoints 3 and 4 so each question has a single clear intent. CP3 Q2 now asks only to recall the three CEI steps (the "why" is already covered by Q1). CP4 Q2 now asks only what happens to trapped ETH, which is derivable from context 🙌

rin-st (approved):
Thank you for updates! Lgtm!

PR #451, opened 2026-04-21
fix: update InsufficientLiquidity error args in IDEX.sol and DEX.t.sol

whoisrosea:
The README tells to declare `InsufficientLiquidity(uint256 available, uint256 required)`, 
but `IDEX.sol` has it without args. Since the Foundry test uses `abi.encodeWithSelector`, 
argument types matter and the test fails.

Fixed `IDEX.sol` error signature and updated `DEX.t.sol` accordingly.


rin-st:
Thank you @whoisrosea !

PR #460, opened 2026-05-22
feat: prediction-markets migration to hh v3

rin-st:
## Summary
- Migrate Hardhat deploy to `rocketh`'s `deployScript` (uses `env.read` for view calls and `env.network.provider.request` for raw RPC).
- Replace `hre.ethers.parseEther` with `viem`'s `parseEther` in the deploy.
- Migrate test to HH v3: `network.create()` in `before` hook, type-only imports from `../types/ethers-contracts/index.js`, cast deploys with `as unknown as`.

## Test plan
- [ ] `yarn deploy` deploys `PredictionMarket` and writes YES/NO token deployment files
- [ ] `yarn hardhat:test` passes


extension/packages/nextjs/components/race/RaceEffects.tsx
```diff
@@ -1,5 +1,6 @@
 "use client";
 
+/* eslint-disable react-hooks/purity */
```
damianmarti, line 3: Why do you need this?

extension/packages/hardhat/deploy/00_deploy_your_contract.ts
```diff
@@ -1,77 +1,57 @@
...
+    const initialProbability = 50;
+    const percentageLocked = 10;
+    const liquidityProvider = deployer;
+    const oracle = deployer;
+
+    const predictionMarket = await env.deploy("PredictionMarket", {
+      account: deployer,
+      artifact: artifacts.PredictionMarket,
+      args: [liquidityProvider, oracle, question, initialTokenValue, initialProbability, percentageLocked],
+      value: initialLiquidity,
```
damianmarti, line 33: The funds are not sent to the contract. 

We have to send the funds after the deployment (but we will need to change the constructor solution from README) or use env.broadcastDeployment to deploy the contract.

damianmarti (commented):
Thanks for the migration!

We have an issue when deploying the contract. The funds are not sent, the same issue that happened on another challenge.

There were type errors in the test file. I pushed a fix using PredictionMarket__factory. Please review https://github.com/scaffold-eth/se-2-challenges/pull/460/changes/3f3f333b78569a8a8d807ef30e48aae00aff8c0f

Everything else is working great!

Thanks!

extension/packages/nextjs/components/race/RaceEffects.tsx
rin-st: Try to uncomment it and you'll see lint errors for every `Math.random()`

extension/packages/hardhat/deploy/00_deploy_your_contract.ts
rin-st: it doesn't work like in dice challenge since in checkpoint 2 we have in constructor this
```
        if (msg.value == 0) revert PredictionMarket__MustProvideETHForInitialLiquidity();
```

so adding value later doesn't work. For now I added patched version of rocket/deploy [adbbcc9](https://github.com/scaffold-eth/se-2-challenges/pull/460/commits/adbbcc9f549f0aa3841e9451cd3996b2b2b8976f) and created the [issue](https://github.com/wighawag/rocketh/issues/40) in rocketh

damianmarti (approved):
Working great now! Thanks!!

extension/packages/hardhat/deploy/00_deploy_your_contract.ts
rin-st: [9b98638](https://github.com/scaffold-eth/se-2-challenges/pull/460/commits/9b98638a3acf48a2539550f4752c59e6787dfc6c) reverted patched yarn version since issue is resolved and now rocket/deploy allows to send value

damianmarti: great! thanks!!

rin-st: Reverting it again since we have `npmMinimalAgeGate: 7d`. Will fix it after the gate age ends

PR #464, opened 2026-05-22
feat: token-vendor migration to hh v3

rin-st:
## Summary
- Migrate both deploys to `rocketh`'s `deployScript`. `01_deploy_vendor.ts` uses `env.get("YourToken")` to look up the prior deployment.
- Use `env.execute` for `transfer` / `transferOwnership` calls; replace `hre.ethers.parseEther` with `viem`'s `parseEther`.
- Migrate test: `network.create()` in `before` (also destructures `networkHelpers`); replace `impersonateAccount` / `stopImpersonatingAccount` from `@nomicfoundation/hardhat-network-helpers` with `networkHelpers.impersonateAccount` / `networkHelpers.stopImpersonatingAccount`.
- Drop the `import hre from "hardhat"` + `const { ethers } = hre` pattern.
- Cast deploys with `as unknown as`.

## Test plan
- [ ] `yarn deploy` deploys `YourToken` (and `Vendor` when `SEND_TOKENS_TO_VENDOR=true`)
- [ ] `yarn hardhat:test` passes (including the EthTransferFailed impersonation test)

### DIFF
--- a/extension/packages/hardhat/test/Vendor.ts
+++ b/extension/packages/hardhat/test/Vendor.ts
@@ -66,7 +66,7 @@ describe("🚩 Challenge: 🏵 Token Vendor 🤖", function () {
       const { deployer, user, yourToken } = await deployYourTokenFixture();
 
       const amount = ethers.parseEther("10");
-      await expect(yourToken.transfer(user.address, amount)).to.not.be.reverted;
+      await expect(yourToken.transfer(user.address, amount)).to.not.be.revert(ethers);
 
       expect(await yourToken.balanceOf(user.address)).to.equal(amount);
       expect(await yourToken.balanceOf(deployer.address)).to.equal(INITIAL_SUPPLY - amount);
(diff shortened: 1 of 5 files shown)


extension/packages/hardhat/test/Vendor.ts
damianmarti: It has to be changed to

```
.to.not.revert(ethers);
```

damianmarti: This too.

damianmarti (approved):
Working great! 

I only added two comments to the tests.

It's good to go after fixing them.

Thanks!!

extension/packages/hardhat/test/Vendor.ts
rin-st: updated, thanks!

rin-st: [f443271](https://github.com/scaffold-eth/se-2-challenges/pull/464/commits/f443271fc1d730cb019a543c6e1a957555e28ab1) updated, thanks!

PR #466, opened 2026-05-22
feat: crowdfunding migration to hh v3

rin-st:
## Summary
- Migrate both deploys to `rocketh`'s `deployScript`. `01_deploy_crowdfund.ts` uses `env.get("FundingRecipient")` (synchronous in rocketh) to look up the prior deployment.
- Migrate test: `network.create()` in `before` destructuring both `ethers` and `provider`; replace `network.provider.send(method, params)` with `provider.request({ method, params })`; cast deploys with `as unknown as`.

## Test plan
- [ ] `yarn deploy` deploys `FundingRecipient` then `CrowdFund` wired to it
- [ ] `yarn hardhat:test` passes (including the `hardhat_setStorageAt` storage-walk test and the time-warped withdrawal tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Before you begin, you need to install the following tools:
 Then download the challenge to your computer and install dependencies by running:
 
 ```sh
-npx create-eth@2.0.18 -e challenge-crowdfunding challenge-crowdfunding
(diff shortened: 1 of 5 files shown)


damianmarti (approved):
Working great! Thanks!

PR #469, opened 2026-05-22
docs: clarify yarn deploy --reset instructions

abhisarxverma:
Hi team! The current text states that `--reset` should be run if you haven't made contract changes. This is confusing for learners because `--reset` is used to force a deployment or wipe local cache state when changes have been made or are stuck. Updated the wording for clarity. This is my first open-source contribution!

Here is how it currently looks on the website:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ yarn start
 
 📱 Open http://localhost:3000 to see the app.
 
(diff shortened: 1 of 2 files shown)


rin-st:
Thanks for noticing it and for your PR @abhisarxverma ! Rephrased a bit and added changes to  readme args file. Merging!

PR #470, opened 2026-06-04
fail ethToToken test when ETH reserves ignore msg.value

fede-rodes:
## Summary

Tightens the checkpoint 4 `ethToToken` tests in Foundry and Hardhat so a wrong ETH reserve calculation no longer passes as long as the user receives some tokens.

## Why change this?

The checkpoint 4 `ethToToken` test only checked that the user received more tokens. That works as a smoke test, but it misses a common mistake: using `address(this).balance` as the ETH reserve without subtracting `msg.value`. In that case `price()` runs against the wrong pool size (for example 6 ETH instead of 5 right after `init`). The user still gets tokens, so the old assertion passed even though the swap math was off. I only caught this when I happened to read the hint after already getting the test green with the wrong implementation.

## What changed

In Foundry and Hardhat, the test now reads the DEX’s ETH and token balances before the swap, computes the expected output with `price()`, and asserts the user’s balance increases by exactly that amount. The DEX ETH balance is checked as `xReserves + xInput` instead of a hardcoded 6 ETH. Hardhat also asserts the `EthToTokenSwap` event includes the expected token amount.

## How to verify

Implement `ethToToken` along these lines (the important part is `ethReserves = address(this).balance - msg.value`), then run checkpoint 4 in both runners:

```solidity
function ethToToken() public payable returns (uint256 tokenOutput) {
    if (msg.value == 0) {
        revert InvalidEthAmount();
    }

    uint256 ethReserves = address(this).balance - msg.value;
    uint256 tokenReserves = token.balanceOf(address(this));

    tokenOutput = price(msg.value, ethReserves, tokenReserves);

    bool success = token.transfer(msg.sender, tokenOutput);
    if (!success) {
        revert EthTransferFailed(msg.sender, tokenOutput);
    }

    emit EthToTokenSwap(msg.sender, tokenOutput, msg.value);
}
```

To confirm the test actually guards this, temporarily remove the `- msg.value` and checkpoint 4 should fail on the balance assertion.

### DIFF
--- a/extension/packages/foundry/test/DEX.t.sol
+++ b/extension/packages/foundry/test/DEX.t.sol
@@ -97,17 +97,21 @@ contract DEXTest is Test {
     function test_Checkpoint4_EthToTokenEmitsAndTransfers() public {
         _initDex();
 
+        uint256 xInput = 1 ether;
+        uint256 xReserves = address(dex).balance;
+        uint256 yReserves = balloons.balanceOf(address(dex));
+        uint256 expectedYOutput = dex.price(xInput, xReserves, yReserves);
+
         uint256 userBalBefore = balloons.balanceOf(user2);
 
         vm.prank(user2);
         vm.recordLogs();
-        dex.ethToToken{ value: 1 ether }();
+        dex.ethToToken{ value: xInput }();
 
         uint256 userBalAfter = balloons.balanceOf(user2);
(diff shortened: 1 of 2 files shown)


rin-st:
Thank you @fede-rodes ! Merged!

